Issue #1, opened 2023-12-22
Zenoh Rust uLink Library: Initial Commit


stevenhartley:
Implement uTransport & RpcClient interfaces

stevenhartley:
Initial commit of the library, work being done on https://github.com/ZettaScaleLabs/uprotocol-java-ulink-zenoh/tree/initial_version

evshary:
Here is the initial implementation for publisher/subscriber & RpcServer/RpcClient
https://github.com/ZettaScaleLabs/uprotocol-rust-ulink-zenoh/pull/2
I'll send PR after updating the examples by using micro-form Uuri.
It would help a lot if there is any example of how to create micro-form Uuri in uProtocol.

stevenhartley:
@evshary ,
I've added an example in the test for Java, Below that generates an Update notification event emitted from uSubscription service:
```
        UUri uri = UUri.newBuilder()
            .setAuthority(UAuthority.newBuilder()
                .setIp(ByteString.copyFrom(InetAddress.getByName("10.0.3.3").getAddress()))
                .build())
            /* Entity information will be pulled from the generated stubs, for now we hardcode it */
            .setEntity(UEntity.newBuilder()
                .setName("core.usubscription")
                .setId(0)
                .setVersionMajor(3)
                .build())
            /* Resource information is pulled from protobuf generated messages */
            .setResource(UResource.newBuilder()
                .setName(Update.getDescriptor().getName())
                .setInstance(Update.Resources.subscriptions.name())
                .setId(Update.Resources.subscriptions.getNumber())
                .build())
            .build()

```
You can get more info (for Java) here: https://github.com/eclipse-uprotocol/uprotocol-java/blob/main/src/main/java/org/eclipse/uprotocol/uri/README.adoc. Please reach out to Daniel (rust) for examples

evshary:
@stevenhartley I have some follow-up questions:

1. According to the example [here](https://github.com/eclipse-uprotocol/uprotocol-java/blob/main/src/main/java/org/eclipse/uprotocol/uri/README.adoc#21-building-an-rpc-method), does that mean users should generate the ID (number) by themselves based on the rules [here](https://github.com/eclipse-uprotocol/uprotocol-spec/blob/main/basics/uri.adoc#42-micro-uris)?
2. Although Zenoh only uses micro URI, users still need to set the name for entity and resource in their application, right? Is it possible for users to set the ID (number) only?

[...]

PR #2, opened 2024-01-19
The initial implementation of up-client-zenoh-rust

evshary:
The PR is the first implementation of up-client-zenoh-rust, which is ported from [uprotocol-rust-ulink-zenoh](https://github.com/ZettaScaleLabs/uprotocol-rust-ulink-zenoh)
The PR is not ready to be merged, because we still need some changes on [up-rust](https://github.com/eclipse-uprotocol/up-rust)

Related to #1

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,7 +40,7 @@ use zenoh::{
 };
 
 pub struct ZenohListener {}
-pub struct ULinkZenoh {
+pub struct UPClientZenoh {
     session: Arc<Session>,
     subscriber_map: Arc<Mutex<HashMap<String, Subscriber<'static, ()>>>>,
     queryable_map: Arc<Mutex<HashMap<String, Queryable<'static, ()>>>>,
@@ -48,17 +48,17 @@ pub struct ULinkZenoh {
     callback_counter: AtomicU64,
 }
 
-impl ULinkZenoh {
+impl UPClientZenoh {
     /// # Errors
     /// Will return `Err` if unable to create Zenoh session
-    pub async fn new(config: Config) -> Result<ULinkZenoh, UStatus> {
+    pub async fn new(config: Config) -> Result<UPClientZenoh, UStatus> {
         let Ok(session) = zenoh::open(config).res().await else {
             return Err(UStatus::fail_with_code(
                 UCode::Internal,
                 "Unable to open Zenoh session",
             ));
         };
-        Ok(ULinkZenoh {
+        Ok(UPClientZenoh {
             session: Arc::new(session),
             subscriber_map: Arc::new(Mutex::new(HashMap::new())),
             queryable_map: Arc::new(Mutex::new(HashMap::new())),
@@ -112,7 +112,7 @@ impl ULinkZenoh {
         };
 
         // Serialized UAttributes into protobuf
-        let priority = ULinkZenoh::map_zenoh_priority(attributes.priority());
+        let priority = UPClientZenoh::map_zenoh_priority(attributes.priority());
         let mut attr = vec![];
         let Ok(()) = attributes.encode(&mut attr) else {
             return Err(UStatus::fail_with_code(
@@ -214,7 +214,7 @@ impl ULinkZenoh {
 }
 
 #[async_trait]
-impl RpcClient for ULinkZenoh {
+impl RpcClient for UPClientZenoh {
     async fn invoke_method(
         &self,
         topic: UUri,
@@ -237,7 +237,7 @@ impl RpcClient for ULinkZenoh {
         }
 
         // Get Zenoh key
-        let Ok(zenoh_key) = ULinkZenoh::to_zenoh_key_string(&topic) else {
+        let Ok(zenoh_key) = UPClientZenoh::to_zenoh_key_string(&topic) else {
             return Err(RpcMapperError::UnexpectedError(String::from(
                 "Unable to transform to Zenoh key",
             )));
@@ -309,7 +309,7 @@ impl RpcClient for ULinkZenoh {
 }
 
 #[async_trait]
-impl RpcServer for ULinkZenoh {
+impl RpcServer for UPClientZenoh {
     async fn register_rpc_listener(
         &self,
         method: UUri,
@@ -320,7 +320,7 @@ impl RpcServer for ULinkZenoh {
             .map_err(|_| UStatus::fail_with_code(UCode::InvalidArgument, "Invalid topic"))?;
 
         // Get Zenoh key
-        let zenoh_key = ULinkZenoh::to_zenoh_key_string(&method)?;
+        let zenoh_key = UPClientZenoh::to_zenoh_key_string(&method)?;
         // Generate listener string for users to delete
         let hashmap_key = format!(
             "{}_{:X}",
@@ -438,7 +438,7 @@ impl RpcServer for ULinkZenoh {
 }
 
 #[async_trait]
-impl UTransport for ULinkZenoh {
+impl UTransport for UPClientZenoh {
     async fn authenticate(&self, _entity: UEntity) -> Result<(), UStatus> {
         // TODO: Not implemented
         Err(UStatus::fail_with_code(
@@ -458,7 +458,7 @@ impl UTransport for ULinkZenoh {
             .map_err(|_| UStatus::fail_with_code(UCode::InvalidArgument, "Invalid topic"))?;
 
         // Get Zenoh key
-        let zenoh_key = ULinkZenoh::to_zenoh_key_string(&topic)?;
+        let zenoh_key = UPClientZenoh::to_zenoh_key_string(&topic)?;
 
         // Check the type of UAttributes (Publish / Request / Response)
         match UMessageType::try_from(attributes.r#type) {
@@ -503,7 +503,7 @@ impl UTransport for ULinkZenoh {
             .map_err(|_| UStatus::fail_with_code(UCode::InvalidArgument, "Invalid topic"))?;
 
         // Get Zenoh key
-        let zenoh_key = ULinkZenoh::to_zenoh_key_string(&topic)?;
+        let zenoh_key = UPClientZenoh::to_zenoh_key_string(&topic)?;
         // Generate listener string for users to delete
         let hashmap_key = format!(
             "{}_{:X}",
@@ -619,7 +619,7 @@ mod tests {
             ..Default::default()
         };
         assert_eq!(
-            ULinkZenoh::to_zenoh_key_string(&uuri).unwrap(),
+            UPClientZenoh::to_zenoh_key_string(&uuri).unwrap(),
             String::from("0100162e04d20100")
         );
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,28 +11,17 @@
 // Contributors:
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
-use async_trait::async_trait;
+pub mod rpc;
+pub mod utransport;
+
 use protobuf::{Enum, Message};
 use std::{
     collections::HashMap,
-    sync::{
-        atomic::{AtomicU64, Ordering},
-        Arc, Mutex,
-    },
-    time::Duration,
+    sync::{atomic::AtomicU64, Arc, Mutex},
 };
 use up_rust::{
-    rpc::{CallOptions, RpcClient, RpcClientResult, RpcMapperError, RpcServer},
-    transport::{datamodel::UTransport, validator::Validators},
-    uprotocol::{
-        Data, UAttributes, UCode, UMessage, UMessageType, UPayload, UPayloadFormat, UPriority,
-        UStatus, UUri, UUID,
-    },
-    uri::{
-        serializer::{MicroUriSerializer, UriSerializer},
-        validator::UriValidator,
-    },
-    uuid::builder::UUIDBuilder,
+    uprotocol::{Data, UAttributes, UCode, UPayload, UPayloadFormat, UPriority, UStatus, UUri},
+    uri::serializer::{MicroUriSerializer, UriSerializer},
 };
 use zenoh::{
     config::Config,
@@ -106,10 +95,10 @@ impl UPClientZenoh {
         UPayloadFormat::from_i32(value)
     }
 
-    // TODO: Transformation between uAttributes and attachment
-    //       Transform all members first and then exclude some we don't need
     fn uattributes_to_attachment(uattributes: &UAttributes) -> anyhow::Result<AttachmentBuilder> {
         let mut attachment = AttachmentBuilder::new();
+        attachment.insert("attr", &uattributes.write_to_bytes()?);
+        /* TODO: We send the whole uattributes directly for the time being and do the benchmark later.
         attachment.insert("id", &uattributes.id.write_to_bytes()?);
         attachment.insert(
             "type_",
@@ -139,11 +128,21 @@ impl UPClientZenoh {
         if let Some(traceparent) = uattributes.traceparent.clone() {
             attachment.insert("traceparent", &traceparent);
         }
+        */
         Ok(attachment)
     }
 
-    // TODO: Same as uattributes_to_attachment
     fn attachment_to_uattributes(attachment: &Attachment) -> anyhow::Result<UAttributes> {
+        let uattributes = UAttributes::parse_from_bytes(
+            attachment
+                .get(&"attr".as_bytes())
+                .ok_or(UStatus::fail_with_code(
+                    UCode::INTERNAL,
+                    "Unable to get uAttributes",
+                ))?
+                .as_slice(),
+        )?;
+        /* TODO: We send the whole uattributes directly for the time being and do the benchmark later.
         let mut uattributes = UAttributes::new();
         if let Some(id) = attachment.get(&"id".as_bytes()) {
             let uuid = UUID::parse_from_bytes(&id)?;
@@ -204,6 +203,7 @@ impl UPClientZenoh {
             let traceparent = traceparent.to_string();
             uattributes.traceparent = Some(traceparent);
         }
+        */
         Ok(uattributes)
     }
 
@@ -310,394 +310,6 @@ impl UPClientZenoh {
             .await
             .map_err(|_| UStatus::fail_with_code(UCode::INTERNAL, "Unable to reply with Zenoh"))?;
 
-        Ok(())
-    }
-}
-
-// TODO: Need to check how to use CallOptions
-#[async_trait]
-impl RpcClient for UPClientZenoh {
-    async fn invoke_method(
-        &self,
-        topic: UUri,
-        payload: UPayload,
-        _options: CallOptions,
-    ) -> RpcClientResult {
-        // Validate UUri
-        UriValidator::validate(&topic)
-            .map_err(|_| RpcMapperError::UnexpectedError(String::from("Wrong UUri")))?;
-
-        // TODO: Comment it and check whether uAttributes is not necessary.
-        //// Validate UAttributes
-        //{
-        //    // TODO: Check why the validator doesn't have Send
-        //    let validator = Validators::Request.validator();
-        //    if let Err(e) = validator.validate(&attributes) {
-        //        return Err(RpcMapperError::UnexpectedError(format!(
-        //            "Wrong UAttributes {e:?}",
-        //        )));
-        //    }
-        //}
-
-        // Get Zenoh key
-        let Ok(zenoh_key) = UPClientZenoh::to_zenoh_key_string(&topic) else {
-            return Err(RpcMapperError::UnexpectedError(String::from(
-                "Unable to transform to Zenoh key",
-            )));
-        };
-
-        // Get the data from UPayload
-        let Some(Data::Value(buf)) = payload.data else {
-            // TODO: Assume we only have Value here, no reference for shared memory
-            return Err(RpcMapperError::InvalidPayload(String::from(
-                "Wrong UPayload",
-            )));
-        };
-
-        // TODO: Check how to generate uAttributes, ex: source, reqid...
-        let mut uattributes = UAttributes::new();
-        uattributes.source = Some(topic.clone()).into();
-        uattributes.sink = Some(topic).into();
-        uattributes.reqid = Some(UUIDBuilder::new().build()).into();
-        let Ok(attachment) = UPClientZenoh::uattributes_to_attachment(&uattributes) else {
-            return Err(RpcMapperError::UnexpectedError(String::from(
-                "Invalid uAttributes",
-            )));
-        };
-
-        let value = Value::new(buf.into()).encoding(Encoding::WithSuffix(
-            KnownEncoding::AppCustom,
-            payload.format.value().to_string().into(),
-        ));
-        // TODO: Query should support .encoding
-        // TODO: Adjust the timeout
-        let getbuilder = self
-            .session
-            .get(&zenoh_key)
-            .with_value(value)
-            .with_attachment(attachment.build())
-            .target(QueryTarget::BestMatching)
-            .timeout(Duration::from_millis(1000));
-
-        // Send the query
-        let Ok(replies) = getbuilder.res().await else {
-            return Err(RpcMapperError::UnexpectedError(String::from(
-                "Error while sending Zenoh query",
-            )));
-        };
-
-        let Ok(reply) = replies.recv_async().await else {
-            return Err(RpcMapperError::UnexpectedError(String::from(
-                "Error while receiving Zenoh reply",
-            )));
-        };
-        match reply.sample {
-            Ok(sample) => {
-                let Some(encoding) = UPClientZenoh::to_upayload_format(&sample.encoding) else {
-                    return Err(RpcMapperError::UnexpectedError(String::from(
-                        "Error while parsing Zenoh encoding",
-                    )));
-                };
-                // TODO: Need to check attributes is correct or not
-                Ok(UMessage {
-                    attributes: Some(uattributes).into(),
-                    payload: Some(UPayload {
-                        length: Some(0),
-                        format: encoding.into(),
-                        data: Some(Data::Value(sample.payload.contiguous().to_vec())),
-                        ..Default::default()
-                    })
-                    .into(),
-                    ..Default::default()
-                })
-            }
-            Err(_) => Err(RpcMapperError::UnexpectedError(String::from(
-                "Error while parsing Zenoh reply",
-            ))),
-        }
-    }
-}
-
-#[async_trait]
-impl RpcServer for UPClientZenoh {
-    async fn register_rpc_listener(
-        &self,
-        method: UUri,
-        listener: Box<dyn Fn(Result<UMessage, UStatus>) + Send + Sync + 'static>,
-    ) -> Result<String, UStatus> {
-        // Do the validation
-        UriValidator::validate(&method)
-            .map_err(|_| UStatus::fail_with_code(UCode::INVALID_ARGUMENT, "Invalid topic"))?;
-
-        // Get Zenoh key
-        let zenoh_key = UPClientZenoh::to_zenoh_key_string(&method)?;
-        // Generate listener string for users to delete
-        let hashmap_key = format!(
-            "{}_{:X}",
-            zenoh_key,
-            self.callback_counter.fetch_add(1, Ordering::SeqCst)
-        );
-
-        let query_map = self.query_map.clone();
-        // Setup callback
-        let callback = move |query: Query| {
-            // Create UAttribute
-            let Some(attachment) = query.attachment() else {
-                listener(Err(UStatus::fail_with_code(
-                    UCode::INTERNAL,
-                    "Unable to get attachment",
-                )));
-                return;
-            };
-            let u_attribute = match UPClientZenoh::attachment_to_uattributes(attachment) {
-                Ok(uattributes) => uattributes,
-                Err(e) => {
-                    log::error!("attachment_to_uattributes error: {:?}", e);
-                    listener(Err(UStatus::fail_with_code(
-                        UCode::INTERNAL,
-                        "Unable to decode attribute",
-                    )));
-                    return;
-                }
-            };
-            // Create UPayload
-            let u_payload = match query.value() {
-                Some(value) => {
-                    let Some(encoding) = UPClientZenoh::to_upayload_format(&value.encoding) else {
-                        listener(Err(UStatus::fail_with_code(
-                            UCode::INTERNAL,
-                            "Unable to get payload encoding",
-                        )));
-                        return;
-                    };
-                    UPayload {
-                        length: Some(0),
-                        format: encoding.into(),
-                        data: Some(Data::Value(value.payload.contiguous().to_vec())),
-                        ..Default::default()
-                    }
-                }
-                None => UPayload {
-                    length: Some(0),
-                    format: UPayloadFormat::UPAYLOAD_FORMAT_UNSPECIFIED.into(),
-                    data: None,
-                    ..Default::default()
-                },
-            };
-            // Create UMessage
-            let msg = UMessage {
-                attributes: Some(u_attribute.clone()).into(),
-                payload: Some(u_payload).into(),
-                ..Default::default()
-            };
-            query_map
-                .lock()
-                .unwrap()
-                .insert(u_attribute.reqid.to_string(), query);
-            listener(Ok(msg));
-        };
-        if let Ok(queryable) = self
-            .session
-            .declare_queryable(&zenoh_key)
-            .callback_mut(callback)
-            .res()
-            .await
-        {
-            self.queryable_map
-                .lock()
-                .unwrap()
-                .insert(hashmap_key.clone(), queryable);
-        } else {
-            return Err(UStatus::fail_with_code(
-                UCode::INTERNAL,
-                "Unable to register callback with Zenoh",
-            ));
-        }
-
-        Ok(hashmap_key)
-    }
-    async fn unregister_rpc_listener(&self, method: UUri, listener: &str) -> Result<(), UStatus> {
-        // Do the validation
-        UriValidator::validate(&method)
-            .map_err(|_| UStatus::fail_with_code(UCode::INVALID_ARGUMENT, "Invalid topic"))?;
-        // TODO: Check whether we still need method or not (Compare method with listener?)
-
-        if self
-            .queryable_map
-            .lock()
-            .unwrap()
-            .remove(listener)
-            .is_none()
-        {
-            return Err(UStatus::fail_with_code(
-                UCode::INVALID_ARGUMENT,
-                "Listener doesn't exist",
-            ));
-        }
-
-        Ok(())
-    }
-}
-
-#[async_trait]
-impl UTransport for UPClientZenoh {
-    async fn send(&self, message: UMessage) -> Result<(), UStatus> {
-        let payload = *message.payload.0.ok_or(UStatus::fail_with_code(
-            UCode::INVALID_ARGUMENT,
-            "Invalid uPayload",
-        ))?;
-        let attributes = *message.attributes.0.ok_or(UStatus::fail_with_code(
-            UCode::INVALID_ARGUMENT,
-            "Invalid uAttributes",
-        ))?;
-        let topic = attributes.clone().sink;
-        // Do the validation
-        UriValidator::validate(&topic)
-            .map_err(|_| UStatus::fail_with_code(UCode::INVALID_ARGUMENT, "Invalid topic"))?;
-
-        // Get Zenoh key
-        let zenoh_key = UPClientZenoh::to_zenoh_key_string(&topic)?;
-
-        // Check the type of UAttributes (Publish / Request / Response)
-        match attributes
-            .type_
-            .enum_value()
-            .map_err(|_| UStatus::fail_with_code(UCode::INTERNAL, "Unable to parse type"))?
-        {
-            UMessageType::UMESSAGE_TYPE_PUBLISH => {
-                Validators::Publish
-                    .validator()
-                    .validate(&attributes)
-                    .map_err(|e| {
-                        UStatus::fail_with_code(
-                            UCode::INVALID_ARGUMENT,
-                            format!("Wrong Response UAttributes {e:?}"),
-                        )
-                    })?;
-                self.send_publish(&zenoh_key, payload, attributes).await
-            }
-            UMessageType::UMESSAGE_TYPE_RESPONSE => {
-                Validators::Response
-                    .validator()
-                    .validate(&attributes)
-                    .map_err(|e| {
-                        UStatus::fail_with_code(
-                            UCode::INVALID_ARGUMENT,
-                            format!("Wrong Response UAttributes {e:?}"),
-                        )
-                    })?;
-                self.send_response(&zenoh_key, payload, attributes).await
-            }
-            _ => Err(UStatus::fail_with_code(
-                UCode::INVALID_ARGUMENT,
-                "Wrong Message type in UAttributes",
-            )),
-        }
-    }
-
-    async fn receive(&self, _topic: UUri) -> Result<UMessage, UStatus> {
-        Err(UStatus::fail_with_code(
-            UCode::UNIMPLEMENTED,
-            "Not implemented",
-        ))
-    }
-
-    async fn register_listener(
-        &self,
-        topic: UUri,
-        listener: Box<dyn Fn(Result<UMessage, UStatus>) + Send + Sync + 'static>,
-    ) -> Result<String, UStatus> {
-        // Do the validation
-        UriValidator::validate(&topic)
-            .map_err(|_| UStatus::fail_with_code(UCode::INVALID_ARGUMENT, "Invalid topic"))?;
-
-        // Get Zenoh key
-        let zenoh_key = UPClientZenoh::to_zenoh_key_string(&topic)?;
-        // Generate listener string for users to delete
-        let hashmap_key = format!(
-            "{}_{:X}",
-            zenoh_key,
-            self.callback_counter.fetch_add(1, Ordering::SeqCst)
-        );
-
-        // Setup callback
-        let callback = move |sample: Sample| {
-            // Create UAttribute
-            let Some(attachment) = sample.attachment() else {
-                listener(Err(UStatus::fail_with_code(
-                    UCode::INTERNAL,
-                    "Unable to get attachment",
-                )));
-                return;
-            };
-            let u_attribute = match UPClientZenoh::attachment_to_uattributes(attachment) {
-                Ok(uattributes) => uattributes,
-                Err(e) => {
-                    log::error!("attachment_to_uattributes error: {:?}", e);
-                    listener(Err(UStatus::fail_with_code(
-                        UCode::INTERNAL,
-                        "Unable to decode attribute",
-                    )));
-                    return;
-                }
-            };
-            // Create UPayload
-            let Some(encoding) = UPClientZenoh::to_upayload_format(&sample.encoding) else {
-                listener(Err(UStatus::fail_with_code(
-                    UCode::INTERNAL,
-                    "Unable to get payload encoding",
-                )));
-                return;
-            };
-            let u_payload = UPayload {
-                length: Some(0),
-                format: encoding.into(),
-                data: Some(Data::Value(sample.payload.contiguous().to_vec())),
-                ..Default::default()
-            };
-            // Create UMessage
-            let msg = UMessage {
-                attributes: Some(u_attribute).into(),
-                payload: Some(u_payload).into(),
-                ..Default::default()
-            };
-            listener(Ok(msg));
-        };
-        if let Ok(subscriber) = self
-            .session
-            .declare_subscriber(&zenoh_key)
-            .callback_mut(callback)
-            .res()
-            .await
-        {
-            self.subscriber_map
-                .lock()
-                .unwrap()
-                .insert(hashmap_key.clone(), subscriber);
-        } else {
-            return Err(UStatus::fail_with_code(
-                UCode::INTERNAL,
-                "Unable to register callback with Zenoh",
-            ));
-        }
-
-        Ok(hashmap_key)
-    }
-
-    async fn unregister_listener(&self, topic: UUri, listener: &str) -> Result<(), UStatus> {
-        // Do the validation
-        UriValidator::validate(&topic)
-            .map_err(|_| UStatus::fail_with_code(UCode::INVALID_ARGUMENT, "Invalid topic"))?;
-        // TODO: Check whether we still need topic or not (Compare topic with listener?)
-
-        if !self.subscriber_map.lock().unwrap().contains_key(listener) {
-            return Err(UStatus::fail_with_code(
-                UCode::INVALID_ARGUMENT,
-                "Listener doesn't exist",
-            ));
-        }
-
-        self.subscriber_map.lock().unwrap().remove(listener);
         Ok(())
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,0 +1,631 @@
+//
+// Copyright (c) 2024 ZettaScale Technology
+//
+// This program and the accompanying materials are made available under the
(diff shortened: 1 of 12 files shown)


evshary:
The PR is dependent to https://github.com/eclipse-uprotocol/up-rust/pull/22

src/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+use zenoh::{
+    config::Config,
+    prelude::{r#async::*, Sample},
+    queryable::{Query, Queryable},
+    sample::AttachmentBuilder,
+    subscriber::Subscriber,
+};
+
+pub struct ZenohListener {}
+pub struct ULinkZenoh {
```
stevenhartley, line 43: ULinkZenoh needs to construct its own rpc response topic that it has to send to the server so that the responding uE knows who sent the request (in the RpcServer response listener).

stevenhartley (changes requested):
Pete found a rather important bug in that we are missing the senders address (source) in the RPC flow. When a requester calls invoke_method() the source address (i.e. the return address) is not passed as it is fixed and can easily be constructed in the uPClient constructor.    when the server receives the request and calls the callback, the source URI in the UMessage needs to be the senders rpc response topic (ex. `//vcu.vin/hartley_app/1/rpc.response`).  right now the response address is not passed in the RPC flow so there is no way to know who sent the request.

src/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+                return;
+            };
+            let u_payload = UPayload {
+                length: Some(0),
+                format: encoding,
+                data: Some(Data::Value(sample.payload.contiguous().to_vec())),
+            };
+            // Create UMessage
+            let msg = UMessage {
+                source: Some(topic.clone()),
```
PLeVasseur, line 558: In a similar fashion to @stevenhartley's [comment over here](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/2/files#r1463406504), we'd need to always get the actual source.

Why? In support of a "special mode" for `UTransport::register_listener()` that can support listening to all remotely addressed CEs that the uStreamer would need.

src/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+            UPriority::UpriorityCs5 => Priority::InteractiveHigh,
+            UPriority::UpriorityCs6 => Priority::RealTime,
+            // If uProtocol prioritiy isn't specified, use CS1(DataLow) by default.
+            // https://github.com/eclipse-uprotocol/uprotocol-spec/blob/main/basics/qos.adoc
+            UPriority::UpriorityUnspecified => Priority::DataLow,
+        }
+    }
+
+    // TODO: We need a standard way in uprotocol-rust to change UUID to String
+    fn uuid_to_string(uuid: &Uuid) -> String {
```
PLeVasseur, line 100: There's a method to do this [over here](https://github.com/eclipse-uprotocol/up-rust/blob/main/src/proto/uprotocol/uuid.rs#L43) defined on `Uuid` that can be used. Might've been missed because we rename `Uuid` => `uproto_Uuid` [internal to this file](https://github.com/eclipse-uprotocol/up-rust/blob/f9163f7bc742d6d59b05b35a03666aaaab3cf80d/src/proto/uprotocol/uuid.rs#L17C31-L17C42). :sweat_smile:

src/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+        Ok(ULinkZenoh {
+            session: Arc::new(session),
+            subscriber_map: Arc::new(Mutex::new(HashMap::new())),
+            queryable_map: Arc::new(Mutex::new(HashMap::new())),
+            query_map: Arc::new(Mutex::new(HashMap::new())),
+            callback_counter: AtomicU64::new(0),
+        })
+    }
+
+    fn to_zenoh_key_string(uri: &UUri) -> Result<String, UStatus> {
```
PLeVasseur, line 70: IMHO it would be a good idea to prepend Zenoh keys specific to uProtocol with e.g. "up/" so that we namespace all the uProtocol communication happening over Zenoh to avoid collisions with any other systems on the same network using Zenoh.

PLeVasseur, line 70: In support of the uStreamer, we'll need some kind of special mode. Let me sketch out the idea here, but let's discuss more in tomorrow's meeting.

In uStreamer we need to be able to be able to subscribe to "any CEs which have a UAuthority filled in, and some valid Remote" so probably we need to modify this bit. I have a bit of a hack over [in a feature branch](https://github.com/PLeVasseur/uprotocol-rust-ulink-zenoh_zettascale/blob/feature/stand-in_for_uprotocol-rust-transport-mqtt/src/lib.rs#L116-L142) I'm using to keep moving on development of uStreamer.

So that'd mean that in:

`UTransport::register_listener()` we need a way to listen to all CEs where:
UAttributes.sink.authority has a remote &
UAttributes.type = UmessageTypeRequest

`RpcServer::register_rpc_listener()` we need a way to listen to all CEs where:
UAttributes.sink.authority has a remote &
UAttributes.type = UmessageTypePublish or UmessageTypeResponse

src/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+}
+
+#[async_trait]
+impl RpcClient for ULinkZenoh {
+    async fn invoke_method(
+        &self,
+        topic: UUri,
+        payload: UPayload,
+        attributes: UAttributes,
+    ) -> RpcClientResult {
```
PLeVasseur, line 225: One thing I ran into when writing a uStreamer implementation is that the RpcServer should be responsible for providing (at a minimum) UAttributes back and RpcClientResult should return in addition to a UPayload, at a minimum the UAttributes. Why? This allows for a uStreamer implementation to properly route a response back over another transport. Let's discuss more in our meeting tomorrow.

src/lib.rs
evshary: Yes, I missed that... Thanks!

src/lib.rs
evshary: I don't think the key will conflict since we're using micro-form uuri.
However, I think it's a good way to let others know the key is from uProtocol instead of simply something unreadable to humans.

evshary:
@PLeVasseur I've renamed the uLink to UPClient now.
Just a reminder in case your uStreamer implementation depends on this PR.

evshary:
Hi @PLeVasseur @stevenhartley 
I created an issue here #3 to discuss the "source" issue separately

src/lib.rs
```diff
@@ -0,0 +1,626 @@
...
+    fn to_zenoh_key_string(uri: &UUri) -> Result<String, UStatus> {
+        let micro_uuri = MicroUriSerializer::serialize(uri).map_err(|_| {
+            UStatus::fail_with_code(
+                UCode::InvalidArgument,
+                "Unable to serialize into micro format",
+            )
+        })?;
+        let micro_zenoh_key: String = micro_uuri
+            .iter()
+            .fold(String::new(), |s, c| s + &format!("{c:02x}"));
```
PLeVasseur, line 79: I've been thinking about this more in the context of usage by a uStreamer in support of [the special mode I mention here](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/2/files#r1463538303).

In that case, in order to make it easier to implement the `register_listener(all_with_UAuthority)` functionality, I think choosing an alternative way to arrange the Zenoh key in that case could be beneficial. IMHO we don't need to have the Zenoh key match the Micro URI, so long as we document how we're doing it.

The Micro URI form [as documented here](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#42-micro-uris) has the UAuthority (if any) appended on the end.

However if I ask to `register_listener(all_with_UAuthority)`, implementing that via a Zenoh key could become challenging in `up-client-zenoh-rust`.

Perhaps in addition to placing `up/` in front, we could move the UAuthority in front like this: `up/[byte representation of UAuthority]/[byte representation of rest of UUri]`.

Someone calling `register_listener(all_with_UAuthority)` could have their Zenoh key then be: `up/*/**` to ensure they got only messages with a UAuthority attached.

WDYT?

src/lib.rs
```diff
@@ -0,0 +1,626 @@
...
+impl UTransport for UPClientZenoh {
+    async fn authenticate(&self, _entity: UEntity) -> Result<(), UStatus> {
+        // TODO: Not implemented
+        Err(UStatus::fail_with_code(
+            UCode::Unimplemented,
+            "Not implemented",
+        ))
+    }
+
+    async fn send(
```
PLeVasseur, line 450: Hey @evshary :wave: 

There are some spec clarifications that will need to be made that have an impact on the implementation of send() and register_listener().

Because of how [uP Dispatchers work](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l2/dispatchers) over uP-L1 only, there's a need for a way to interact through up-client-zenoh-rust using only send() and register_listener(all_remote) that can intelligently beneath the hood operate with all of UMessageType::UMessageTypeRequest, UMessageType::UMessageTypeResponse, UMessageType::UMessageTypePublish.

I was looking at a couple of sequence diagrams last week with @stevenhartley that explained the rationale. @stevenhartley  -- could we open a PR to uc-spec with those in them?

Essentially the use case of uStreamer implies that it should call `UTransport::register_listener(special_mode_for_all_UAuthority, all_remote_callback)` and `all_remote_callback` should be triggered for all CEs whether UMessageType::UMessageTypeRequest, UMessageType::UMessageTypeResponse, or UMessageType::UMessageTypePublish

On the other end, since uStreamer talks over uP-L1 only, that means it should always use `UTransport::send(uuri, payload, attributes)` and the transport implementation in up-client-zenoh-rust should send appropriately whether UMessageType::UMessageTypeRequest, UMessageType::UMessageTypeResponse, or UMessageType::UMessageTypePublish. So in the Zenoh case where a pub/sub model is used for UMessageTypePublish and Queryables are used for UMessageTypeRequest & UMessageTypeResponse there'd have to be some thought given there on how to implement that.

If any questions / concerns we can discuss here or in this week's meeting.

src/lib.rs
PLeVasseur: Old comment. Refer to [this one instead](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/2/files#r1469865713) that fleshes the idea out more.

src/lib.rs
PLeVasseur: There's a [spec update](https://github.com/eclipse-uprotocol/up-spec/issues/61) which has invoke_method return a UMessage instead. We need to update up-rust's invoke_method signature before you could make that change tho.

src/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+                }
+                None => UPayload {
+                    length: Some(0),
+                    format: UPayloadFormat::UpayloadFormatUnspecified as i32,
+                    data: None,
+                },
+            };
+            // Create UMessage
+            let msg = UMessage {
+                source: Some(method.clone()),
```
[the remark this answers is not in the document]
PLeVasseur: Following back up on this, @stevenhartley showed me [how this is accomplished in up-client-android-java](https://github.com/eclipse-uprotocol/up-client-android-java/pull/3/files#diff-fddb352c27884216476f4ee1bc524c0a823c41de369dad3f1cb038e4c4429643R214) last week, so I'll link it here in case you want to reference it.

src/lib.rs
evshary: @PLeVasseur I don't quite get the idea why we need to separate UAuthourity in Zenoh key. Can't we just use `up/**`?

PLeVasseur: Oh, because [not every Micro Uri must contain a UAuthority](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#42-micro-uris):

image

And the order that the bytes are written would normally put the bytes for UAuthority at the end.

src/lib.rs
evshary: After viewing the flow chart mentioned in slack, I understand more now.
Let me put the diagram here for discussion.
image

It's a little bit different from my initial thought.

1. While calling `invoke_method` on uPClientFoo side, I thought uStreamer will also call `invoke_method` on uPClientBar side, but it seems like we call the `send` directly to RpcListener on uPClientBar side.
2. While we call registerRpcListener (UUri: /uService/rpc.myMethod) on uPClientBar side, I thought we'll create the corresponding registerRpcListener (UUri: /uService/rpc.myMethod)  on uPClientFoo side to wait for `invoke_method` on uPClientFoo side. Then we don't need to listen to all the UUri (up/**).

To be honest, I don't have a strong preference on the uStreamer implementations, and UPClientZenoh should be able to support them no matter which implementation we choose.
It's great to know that I have a better understanding on the design now.

src/lib.rs
evshary: I see, so you want to only listen to the Uuri with remote UAuthority and don't need to filter it in the `register_listener`. It seems OK to me. Maybe we can raise the discussion in tomorrow's meeting in case we miss anything.

src/lib.rs
PLeVasseur: Yup, I'll expand a little bit to make sure we're on the same page:

1. Correct, as we're bridging transports and don't want the implementation details of a given up-client-foo-lang to leak into the uStreamer, the uStreamer needs to be a little "dumb" and interact through these simpler uP-L1 APIs.
2. As you can see (think we need to update the diagram's uStreamer UUri), the uStreamer would be listening in on `up/*/**` to catch _any_ CEs originating externally from the Bar transport. The uStreamer would then catch that `UMessage`, and call `send()` on the up-client-foo-lang and under the hood, up-client-foo-lang would then do the appropriate action for the UMessageType (Request, Response, Publish).

Does that make sense?

src/lib.rs
```diff
@@ -0,0 +1,626 @@
...
+            payload.format.to_string().into(),
+        ));
+        // TODO: Query should support .encoding
+        // TODO: Adjust the timeout
+        let getbuilder = self
+            .session
+            .get(&zenoh_key)
+            .with_value(value)
+            .with_attachment(attachment.build())
+            .target(QueryTarget::BestMatching)
```
PLeVasseur, line 276: @evshary -- seems this is part of what Luca was referring to in the meeting today.

Let's work with a specific scenario
* `Device1` and `Device2` both speak Zenoh as their "host transport"
* `Device1` and `Device2` each have a uStreamer with an instance of `up-client-zenoh-rust` running with `register_listener(specific_remote_UAuthority_either_id_or_ip, callback)`
  * `Device1`: within uStreamer: `up_client_zenoh_rust.register_listener(UAuthority_pointed_to_Device1, zenoh_remote_callback)`
  * `Device1`: within uApp: `up_client_zenoh_rust.register_rpc_listener(specific_uuri, callback)`

Situation:
From `Device2` there's an `invoke_method()` which contains a UAuthority which points to `Device1`

Question:
1. Given that both uApp and uStreamer on `Device1` would under-the-hood register on the query, but uApp's Zenoh KeyExpr is _more specific_ and the uStreamer one  _includes wildcards_ (e.g. `/up/<Device2-UAuthority>/**`) which `QueryTarget` would be chosen -- the uApp one or the uStreamer one?

To be clear -- the desire is the _more specific_ one, i.e. the uApp to be chosen if it's found.

As a follow-up question -- 

2. In the case the Queryable on `Device1` cannot be found, e.g. the case @stevenhartley showed where there is a gateway in between and no direct communication is possible between `Device1` and `Device2`, but on the gateway there is a `up_client_zenoh_rust.register_listener(UAuthority_pointed_to_Device2, zenoh_remote_callback)`, would the `zenoh_remote_callback` from the gateway be seen and triggered?

evshary: @PLeVasseur 
Let me use Zenoh's way to answer your questions first.

About your first question:
The query target can be set in 3 different modes (BEST_MATCHING, ALL_COMPLETE, ALL)
https://zenoh-c.readthedocs.io/en/latest/api.html#c.z_query_target_t
Assume we have two queryables with `complete` is true (register to `up/uauth/**` and `up/uauth/app` respectively)
Now we want to send query with `up/uauth/app`
* query with `BEST_MATCHING`: Only the nearest one will reply. Might be `up/uauth/**` or `up/uauth/app`
* query with `ALL_COMPLETE` or `ALL`: Both `up/uauth/**` and `up/uauth/app` will reply

If I understand correctly, you want only `up/uauth/app` to reply the query. If there is no `up/uauth/app`, then `up/uauth/**` will take place. Am I right?
If this is your case, maybe @Mallets can give us some suggestions. I'm not pretty sure whether there is a good way to achieve it.

For the second question, the answer is yes.
In fact, if Device1, Device2 and Gateway are all using Zenoh, Device2 can send query to the queryable in Device1 directly. Therefore, maybe in Zenoh case, Gateway can be just a dummy node running in Zenoh peer mode.

Actually, I also had some questions yesterday.
1. If there are two devices both speak Zenoh and can connect to each other directly, do we still need uStreamer to bridge them?
2. If there are two devices (One speak protocolFoo and one speak protocolBar), should both devices need uStreamers?
```
uApp(protocolFoo) ---> uStreamer ---> uStreamer ---> uApp(protocolBar)
 |----------Device 1-----------|      |----------Device 2-----------|
```
Or maybe we just need one uStreamer
```
uApp(protocolFoo) ---> uStreamer ---> uApp(protocolBar)
 |----------Device 1-----------|      |----Device 2----|
```
3. Maybe the most important one. Let me take your scenario as an example.
In your case, while Device2 calls `invoke_method`, uStreamer needs `register_listener` and forward the request to Device1.
However, in the current implementation, `register_listener` can't receive the message from `invoke_method`. We use query & queryable to implement `invoke_method` and `register_rpc_listener`. Besides, we use put & subscribe to implement `send` and `register_listener`. They are different mechanisms.
Maybe I misunderstand your design, but if not, maybe we need to use put and subscribe to implement `invoke_method` and `register_rpc_listener`.

PLeVasseur: Hey @evshary, I'll try to break our discussion down into parts to answer.

> If I understand correctly, you want only `up/uauth/app` to reply the query. If there is no `up/uauth/app`, then `up/uauth/**` will take place. Am I right? If this is your case, maybe @Mallets can give us some suggestions. I'm not pretty sure whether there is a good way to achieve it.

Yup, seems like a not-unreasonable thing to want to accomplish. Favor the fully-complete-no-wildcard key expression. Would be great to hear more on how that can be accomplished.

> For the second question, the answer is yes.

Gotcha. 

> In fact, if Device1, Device2 and Gateway are all using Zenoh, Device2 can send query to the queryable in Device1 directly. Therefore, maybe in Zenoh case, Gateway can be just a dummy node running in Zenoh peer mode.

I'd have to defer to @stevenhartley on how the gateway would be set up, but I think it'd be running a single other transport. So what you suggest may not be possible in that case.

> Actually, I also had some questions yesterday.
> 
> 1. If there are two devices both speak Zenoh and can connect to each other directly, do we still need uStreamer to bridge them?
> 2. If there are two devices (One speak protocolFoo and one speak protocolBar), should both devices need uStreamers?
> 
> ```
> uApp(protocolFoo) ---> uStreamer ---> uStreamer ---> uApp(protocolBar)
>  |----------Device 1-----------|      |----------Device 2-----------|
> ```
> 
> Or maybe we just need one uStreamer
> 
> ```
> uApp(protocolFoo) ---> uStreamer ---> uApp(protocolBar)
>  |----------Device 1-----------|      |----Device 2----|
> ```

Exactly where I was trying to go yesterday :slightly_smiling_face: I guess since `Device1` and `Device2` _may_ have the same host transport, but they also interact with other devices which only support a third `Baz` transport, they'd have to have uStreamers. (An example of this would be if `Device1` and `Device2` also needed to speak over SOME/IP.)

That means in the case though, of sending from `Device1` speaking Zenoh as host transport to `Device2` speaking Zenoh as host transport _for a CE intended for `Device1`_, the uStreamer in `Device1` would be able to simple drop the queryable. Or at least, that's what I'm hoping is possible. And then let the `uApp` on `Device1` service the Zenoh queryable.

That's where I was going with having the query of invoke_method() favor the _most specific_ key expression in my question up above :point_up_2: 

> 3. Maybe the most important one. Let me take your scenario as an example.
>    In your case, while Device2 calls `invoke_method`, uStreamer needs `register_listener` and forward the request to Device1.
>    However, in the current implementation, `register_listener` can't receive the message from `invoke_method`. We use query & queryable to implement `invoke_method` and `register_rpc_listener`. Besides, we use put & subscribe to implement `send` and `register_listener`. They are different mechanisms.
>    Maybe I misunderstand your design, but if not, maybe we need to use put and subscribe to implement `invoke_method` and `register_rpc_listener`.

Right, so I think this is what Luca was saying yesterday, there's a few changes that'd need to be made I think. Take this as a strawman proposal and feel free to implement in a way that makes sense. I'm certainly not a Zenoh expert.

1. Modify uP-L1's `UTransport::send()` impl to read the `UAttributes.type` to then, under the hood, act appropriately
   * If Publish => as you do today
   * If Response => as you do today
   * If Request => dispatch to a new function which can reach the Zenoh queryable setup by uP-L1 helper's `RpcServer::register_rpc_listener()` _or_ (and this is key) the uP-L1 `UTransport::register_listener()` (more on this in next section)
2. Modify uP-L1's `UTransport::register_listener()` impl to have a "normal mode" and a "uStreamer mode" for lack of a better term. What would that look like?
   * If `UTransport::register_listener(uuri, specific_authority_callback)`'s `uuri` matches some condition, then we'd create a special Zenoh key expression that matches all CEs coming from a specific UAuthority. How would we do that?
   * Let's say for now as a proposal that UUri would look like this:
     * ```rust
       UUri {
           authority: {
               name: <doesnt-matter>
               // either ip or id... depending on which is filled
               ip: <some-specific-ip>,  
               id: <some-specific-id>
           },
           entity: {
             name: "*",
             ..Default::default() // i.e. rest of fields don't matter, set to None
           },
           resource: {
             name: "*",
             ..Default::default() // i.e. rest of fields don't matter, set to None
           }
       }
       ```
   * Then if `register_listener(uuri, specific_authority_callback)` saw that specific kind of UUri, it would both:
     * Set up `specific_authority_callback` to trigger on a Zenoh subscription as per normal, _and in addition_
     * Set up `specific_authority_callback` to trigger on a Zenoh query to `up/<some-specific-ip-or-some-specific-id>/**`
     * (Potentially? This is a bit outside of my knowledge...) Set up `specific_authority_callback` to trigger on the case of receiving a Response type request. I'm not sure how that'd work with the Zenoh queryables and how `invoke_method()` and `register_rpc_listener()` are implemented currently
   * so that the uStreamer could operate with only the uP-L1 layer and push the implementation details on how to accomplish this down into each up-client-foo-lang

Hope that makes more sense, but honestly text is sometimes limiting haha. I'd be happy to jump on a Teams call to discuss further and whiteboard to make sure we're on the same page. Lemme know :slightly_smiling_face:

evshary: Hi @PLeVasseur 

> That means in the case though, of sending from Device1 speaking Zenoh as host transport to Device2 speaking Zenoh as host transport for a CE intended for Device1, the uStreamer in Device1 would be able to simple drop the queryable. Or at least, that's what I'm hoping is possible. And then let the uApp on Device1 service the Zenoh queryable.

I guess uStreamer doesn't support routing messages to another uStreamer, or it might cause some routing issues. (Maybe I'm wrong)
That means uStreamer can know the protocol of both two sides.
Maybe we can make uStreamer ignore the message if it knows both sides are the same protocol.

> Right, so I think this is what Luca was saying yesterday, there's a few changes that'd need to be made I think. Take this as a strawman proposal and feel free to implement in a way that makes sense. I'm certainly not a Zenoh expert.

Good proposal! I'll go through the details and see whether there is another way to fulfill the requirement.
Glad to know that we're on the same page with the issue, and that's why I wrote down the question to double-check with you.

> Hope that makes more sense, but honestly text is sometimes limiting haha. I'd be happy to jump on a Teams call to discuss further and whiteboard to make sure we're on the same page. Lemme know 🙂

Your text is really clear to me now! We can have a discussion on Teams later if there is still something unclear while working on the implementation. Thanks for your clarification.

stevenhartley: It should only get to the streamer when we cross the protocol boundary, otherwise it can stay within the current technology and address space.

evshary: Hi @PLeVasseur 
Sorry for the late reply. Thank you for the nice diagram, and I think Luca has answered your questions on the diagram.
We need to store the query while receiving requests and the RPC callback can respond with the stored query.
Here is [the diagram](https://drive.google.com/file/d/1oFoiqwnLJR3V9sBZqb2tu3YkDwvroPrG/view?usp=sharing) I shared yesterday for your reference.
I'm also happy to have a meeting with you for the details, but I'm off until next Wednesday because of the Lunar New Year.
If you're not in hurry, we can discuss it after then.
uProtocol with Zenoh Design

PLeVasseur: > @PLeVasseur let me attempt to explain a bit better how queries work in Zenoh.
> ...
> Let me know if it makes sense. I'm also happy to jump in an ad-hoc meeting to go through the details.

Great, I think I get it. I updated the diagram [here](https://drive.google.com/file/d/1CPyBC5ZZ4eZwB4NpoEayersPUY3qg0T2/view?usp=sharing) to better capture what is going on then.

Seems like the `up-client-zenoh-rust` would do some book-keeping under the hood to allow us when calling `send()` with a `Response` type, it'd reference the held Zenoh Query we got from the original `register_listener()` and then Reply on it. Am I understanding that correctly?
 * This is the `red` flow on the diagram

On the flip side, when there's an incoming Request, we can use `send()` and it'll dispatch under the hood to `send_request()`. Then the `register_listener()` in the uStreamer would be able to "catch" the reply that's coming back. Is that correct?
 * This is the `green` flow on the diagram

> Hi @PLeVasseur
> ...
> Here is [the diagram](https://drive.google.com/file/d/1oFoiqwnLJR3V9sBZqb2tu3YkDwvroPrG/view?usp=sharing) I shared yesterday for your reference.

Probably I'm just not reading the diagram correctly, but I wanted to ensure that I'm understanding this part:
image

So for every use case other than the uStreamer one where we specify a special UUri which says "give me all remote topics from UAuthority such and such", an app would use either `register_listener` or `register_rpc_listener` as appropriate.

And to confirm: the way you drew the diagram is just meant to mean that a `Receiver` would do _either_ `register_rpc_listener` _or_ `register_listener` and that in the normal case of using `register_listener` on a specific (non-wildcarded) UUri, they would only get Zenoh Subscriptions, and no Zenoh Queryables. Is that right?

Mallets: > > @PLeVasseur let me attempt to explain a bit better how queries work in Zenoh.
> > ...
> > Let me know if it makes sense. I'm also happy to jump in an ad-hoc meeting to go through the details.
> 
> Great, I think I get it. I updated the diagram [here](https://drive.google.com/file/d/1CPyBC5ZZ4eZwB4NpoEayersPUY3qg0T2/view?usp=sharing) to better capture what is going on then.

It looks good to me.

> 
> Seems like the `up-client-zenoh-rust` would do some book-keeping under the hood to allow us when calling `send()` with a `Response` type, it'd reference the held Zenoh Query we got from the original `register_listener()` and then Reply on it. Am I understanding that correctly?
> 
> * This is the `red` flow on the diagram

Correct. 

> 
> On the flip side, when there's an incoming Request, we can use `send()` and it'll dispatch under the hood to `send_request()`. Then the `register_listener()` in the uStreamer would be able to "catch" the reply that's coming back. Is that correct?
> 
> * This is the `green` flow on the diagram

Correct.

> btw @evshary -- from my point of view, we're reaching the "send messages event horizon" and if there are still some things unclear we should tackle them with an online discussion when you're back to make things go faster. We can ask to have this added to the agenda for that week's uProtocol Zenoh meeting.

I'd like to participate as well.

PLeVasseur: > I'd like to participate as well.

I spoke with @stevenhartley offline about giving this agenda item a little room to breathe on our next up-zenoh meeting next Wednesday. Let's tackle there if possible, failing that we could shoot for something offline :+1:

evshary: Hi @PLeVasseur

> So for every use case other than the uStreamer one where we specify a special UUri which says "give me all remote topics from UAuthority such and such", an app would use either register_listener or register_rpc_listener as appropriate.

Yes. This is my understanding.

> And to confirm: the way you drew the diagram is just meant to mean that a Receiver would do either register_rpc_listener or register_listener and that in the normal case of using register_listener on a specific (non-wildcarded) UUri, they would only get Zenoh Subscriptions, and no Zenoh Queryables. Is that right?

According to the discussion last time, Steven said that we can distinguish whether this is an RPC request based on the format UUri. I think we can choose to create either Zenoh Subscriptions or Zenoh Queryables in the register_listener. So, it's yes to your question.

> btw @evshary -- from my point of view, we're reaching the "send messages event horizon" and if there are still some things unclear we should tackle them with an online discussion when you're back to make things go faster. We can ask to have this added to the agenda for that week's uProtocol Zenoh meeting.

Sorry for missing your messages here. I was just back yesterday. I'm happy to discuss it in the next uProtocol Zenoh meeting, or we can have another meeting if necessary.

PLeVasseur: Hey @evshary -- I'm a little unclear on if we're on the same page. Let's bring it to the weekly uProtocol Zenoh meeting.

I'm sensing there may be one, but it's a bit hard to tease out without chatting.

evshary: > I'm sensing that I'm missing something, so best we clear it up through a chat.
Sure. It's also ok if you want to have the chat before the weekly meeting.

> We'd need this to listen to Publish, Request, and Response messages being transmitted over the given UAuthority.
Yes, you're right. What I'm talking about is that we can use some API like `validateRpcMethod` to check whether the URI is for RPC or not. If this is for RPC, then we can use queryable. If not, then keep it in the subscriber way.
https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#3-validators

evshary: OK, I think I get what you mean. You want to have both queryable and subscriber while working in the uStreamer.
Then maybe we need to have some exceptions in register_listener. For example, while the UURI is for `*`, it'll have both queryable and subscriber.

PLeVasseur: Yup! I think we're almost there.

It should be if the UUri's UAuthority is set to something concrete (no wildcards), but the other two fields of UEntity and UResource are wildcards (or empty could work too), then we'd have a "special" uStreamer mode where all messages, whether Request, Response, or Publish are routed to the caller of that register_listener().

IIRC @stevenhartley  was saying we're not doing "default gateway" routing, so that would mean we shouldn't support the UAuthority = "*" case. Feel free to jump in if I mischaracterized.

Hope makes sense.

PLeVasseur: Hey @evshary :wave: 

Slight revision regarding how the `UUri` should be formatted to get `uStreamer`-like behavior on the callback to get all of: `Request`, `Response`, and `Publish` type of messages:

```rust
UUri {
    authority: {
        name: <doesnt-matter-not-used>,
        number: <some-specific-ip-or-id>,
    },
    ..Default::default()
}
```

Or in other words, only the `UUri.UAuthority.number` matters.

The `register_client()` call would then have the Zenoh key expression be we are registering the callback on (for all of `Request`, `Response`, and `Publish`):

`up/<some-specific-ip-or-id-in-2-digit-hex-bytes>/*`

WDYT? Does that make sense?

evshary: Hi @PLeVasseur 
Makes sense to me. I guess the default of entity and resource is empty, which should not be used by the normal UURI. Then we can use it as a special way to listen to both put and query.
Let's take this way first and see if it works or not.

evshary:
@PLeVasseur Just want to let you know I've already updated the uTransport(`send` and `register_listener`). I'll work on refactoring `invoke_method` with `send` and `register_listener` then. If the uStreamer implementation is urgent, maybe you can use the branch (or we can merge it) first. If not, maybe you can wait for me to implement `invoke_method` and add some tests to ensure there are no hidden issues.

src/lib.rs
```diff
@@ -55,18 +62,30 @@ impl UPClientZenoh {
...
+    // The UURI format should be "up/<UAuthority id or ip>/<the rest of UUri>"
     fn to_zenoh_key_string(uri: &UUri) -> Result<String, UStatus> {
         let micro_uuri = MicroUriSerializer::serialize(uri).map_err(|_| {
             UStatus::fail_with_code(
                 UCode::INVALID_ARGUMENT,
                 "Unable to serialize into micro format",
             )
         })?;
-        let micro_zenoh_key: String = micro_uuri
+        let mut micro_zenoh_key = String::from("up/");
```
PLeVasseur, line 78: Hi @evshary -- I am not seeing a "special path" here through which a UUri is checked for the format I outlined [in this comment](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/2#discussion_r1497721355) to be able to avoid the MicroUriSerializer failing.

In my mind, this would involve adding the ability to be able to serialize the UAuthority alone. I think Daniel was planning to make the serialization and deserialization more fine grained and implemented per field in the UUri struct, so it'd be good for you to [speak up on that issue](https://github.com/eclipse-uprotocol/up-rust/issues/40#issuecomment-1961013890) with this requirement.

src/utransport.rs
```diff
@@ -166,20 +326,258 @@ impl UTransport for UPClientZenoh {
...
+    async fn receive(&self, _topic: UUri) -> Result<UMessage, UStatus> {
+        Err(UStatus::fail_with_code(
+            UCode::UNIMPLEMENTED,
+            "Not implemented",
+        ))
+    }
+
+    async fn register_listener(
+        &self,
+        topic: UUri,
```
PLeVasseur, line 505: I'm also not seeing in here special logic to allow us to register for _all three_ of Request, Response, and Publish message types of the UUri matches the format [we discussed here](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/2#discussion_r1474661323).

Is it understood that it is a requirement of the uStreamer on all `up-client-foo-rust` to be able to register to all three types of messages transparently via `UTransport::register_listener()` passing that UUri we agreed upon with only UAuthority filled out?

PLeVasseur: Reading the a little further on how you implemented `UTransport::send()`, with it dispatching to the three types under the hood and how `send_request()` is implemented, I think the requirement could be met by having a special path through this function that checks if we have only UAuthority filled out, and if so, does the contents of all three branches of the if/else if/else below.

Would still have to have a way to have the UAuthority alone serialized, which as I [wrote in this comment](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/2/commits/689bb50c22271f1f5db2aab9d005824aec865d19#r1501858724) should entail that work being pushed into `up-rust` so that it is done consistently across all `up-client-foo-rust`.

Does that make sense?

PLeVasseur: Additionally, it should _not_ be possible to register for Response or Request type messages through `UTransport::register_listener()` at all unless we are in the uStreamer use-case.

So roughly speaking it should probably be:
* Is the UUri the special uStreamer case?
   * Yes => run all three below branches to register for all of Publish, Request, Response
   * No => validate it's a good one to register for Publish messages

Does that make sense?

src/lib.rs
evshary: Hi @PLeVasseur Apologize for missing the logic of creating "special path" here. I will add it.

About the serialization, [the current implementation](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/2/files/1948b92d038d6f01073d687fa00cf9cabc25b6fc..689bb50c22271f1f5db2aab9d005824aec865d19#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R81) extracts the UAUTHORITY_ADDRESS from the serialized Micro URI directly. It's not hard to do so, since UAuthority is the last four bytes in Mirco URI.
https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#422-remote-micro-uris

src/utransport.rs
evshary: > I'm also not seeing in here special logic to allow us to register for all three of Request, Response, and Publish message types of the UUri matches the format https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/2#discussion_r1474661323.

My bad. I'll add the special logic here.

> I think the requirement could be met by having a special path through this function that checks if we have only UAuthority filled out, and if so, does the contents of all three branches of the if/else if/else below.

Agree. But I think we only need to register Publish & Request, while listening to Response is only needed by the side which sends the request.

> Additionally, it should not be possible to register for Response or Request type messages through UTransport::register_listener() at all unless we are in the uStreamer use-case.

I don't think we should do that since `invoke_method` and `register_rpc_listener()` also call `send` and `register_listner()`. What in my mind is that everyone can use `send` and `register_listner()`, but they can simplify the process in the RPC scenario if they use `invoke_method` and `register_rpc_listener()`.
Let's talk about this later in today meeting.

src/lib.rs
evshary: OK, I got what you're talking about. We need the independent UAuthority serialization if we have the special UUri.
I'll raise the issue and now I'll create a workaround function to do the serialization.

src/utransport.rs
PLeVasseur: > Agree. But I think we only need to register Publish & Request, while listening to Response is only needed by the side which sends the request.

The uStreamer will be sending Requests as well though. If the uStreamer gets a Request over sommR, and needs to "forward" that request onto Zenoh, the uStreamer will need a way to listen for the Response to send it back over sommR.

> Let's talk about this later in today meeting.

Sure, sounds good

src/utransport.rs
```diff
@@ -0,0 +1,647 @@
...
+            .ok_or(UStatus::fail_with_code(
+                UCode::INVALID_ARGUMENT,
+                "Lack of source address",
+            ))?
+            .to_string();
+        let resp_callback = self
+            .rpc_callback_map
+            .lock()
+            .unwrap()
+            .get(&resp_listener_str)
```
PLeVasseur, line 122: Making a link

src/utransport.rs
```diff
@@ -0,0 +1,647 @@
...
+            // RPC response
+            let mut listener_str = format!(
+                "{}_{:X}",
+                String::from("up/") + &UPClientZenoh::get_uauth_from_uuri(&topic)? + "/**",
+                self.callback_counter.fetch_add(1, Ordering::SeqCst)
+            );
+            self.rpc_callback_map
+                .lock()
+                .unwrap()
+                .insert(listener_str.clone(), listener.clone());
```
PLeVasseur, line 532: Yeah, this makes sense. I think this should work for the uStreamer use-case. :+1: 

I see you handled the `unregister_listener()` nicely too, so we can provide the concatenated string back as-is and you'll break it and unregister each part. Nice.

src/utransport.rs
```diff
@@ -0,0 +1,647 @@
...
+    }
+
+    async fn register_listener(
+        &self,
+        topic: UUri,
+        listener: Box<dyn Fn(Result<UMessage, UStatus>) + Send + Sync + 'static>,
+    ) -> Result<String, UStatus> {
+        let listener = Arc::new(listener);
+        if topic.authority.is_some() && topic.entity.is_none() && topic.resource.is_none() {
+            // This is special UUri which means we need to register both listeners
```
PLeVasseur, line 522:
```rust
// This is special UUri which means we need to register both listeners
```
=>
```rust
// This is special UUri which means we need to register for all of Publish, Request, and Response
```

src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+            subscriber_map: Arc::new(Mutex::new(HashMap::new())),
+            queryable_map: Arc::new(Mutex::new(HashMap::new())),
+            query_map: Arc::new(Mutex::new(HashMap::new())),
+            rpc_callback_map: Arc::new(Mutex::new(HashMap::new())),
+            callback_counter: AtomicU64::new(0),
+        })
+    }
+
+    // TODO: Workaround function. Should be added in up-rust
+    fn get_uauth_from_uuri(uri: &UUri) -> Result<String, UStatus> {
```
PLeVasseur, line 72: Would be good to review [Daniel's PR](https://github.com/eclipse-uprotocol/up-rust/pull/54) to make sure this fits what you'll need.

PLeVasseur (approved):
Aside from a minor comment nit, this looks good to me so that I can start integrating with it! :slightly_smiling_face:

src/utransport.rs
evshary: Oops 😆

src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+    prelude::r#async::*,
+    queryable::{Query, Queryable},
+    sample::{Attachment, AttachmentBuilder},
+    subscriber::Subscriber,
+};
+
+pub type UtransportListener = Box<dyn Fn(Result<UMessage, UStatus>) + Send + Sync + 'static>;
+
+pub struct ZenohListener {}
+pub struct UPClientZenoh {
```
PLeVasseur, line 38: Making a link

src/utransport.rs
```diff
@@ -0,0 +1,650 @@
...
+                "Unable to register callback with Zenoh",
+            ));
+        }
+
+        Ok(hashmap_key)
+    }
+}
+
+#[async_trait]
+impl UTransport for UPClientZenoh {
```
PLeVasseur, line 439: Making a link

src/lib.rs
```diff
@@ -0,0 +1,736 @@
...
+        attachment.insert(
+            "type_",
+            &uattributes
+                .type_
+                .enum_value()
+                .map_err(|_| UStatus::fail_with_code(UCode::INTERNAL, "Unable to parse type"))?
+                .to_type_string(),
+        );
+        attachment.insert("source", &uattributes.source.write_to_bytes()?);
+        attachment.insert("sink", &uattributes.sink.write_to_bytes()?);
```
sophokles73, line 164: not every message has/needs a *sink*, so I guess this should be done conditionally only

src/lib.rs
```diff
@@ -0,0 +1,736 @@
...
+        if let Some(ttl) = uattributes.ttl {
+            attachment.insert("ttl", &ttl.to_string());
+        }
+        if let Some(plevel) = uattributes.permission_level {
+            attachment.insert("permission_level", &plevel.to_string());
+        }
+        if let Some(commstatus) = uattributes.commstatus {
+            attachment.insert("commstatus", &commstatus.to_string());
+        }
+        attachment.insert("reqid", &uattributes.reqid.write_to_bytes()?);
```
sophokles73, line 176: only response messages contain a *reqid*, so I guess this should be done conditionally only

src/lib.rs
```diff
@@ -0,0 +1,736 @@
...
+            &uattributes
+                .type_
+                .enum_value()
+                .map_err(|_| UStatus::fail_with_code(UCode::INTERNAL, "Unable to parse type"))?
+                .to_type_string(),
+        );
+        attachment.insert("source", &uattributes.source.write_to_bytes()?);
+        attachment.insert("sink", &uattributes.sink.write_to_bytes()?);
+        // TODO: Check whether request & response need priority or not
+        attachment.insert("priority", &uattributes.priority.value().to_string());
```
sophokles73, line 166: I have created https://github.com/eclipse-uprotocol/up-spec/issues/71 in order to shed some light on this ...

src/lib.rs
```diff
@@ -0,0 +1,736 @@
...
+        attributes: UAttributes,
+    ) -> Result<(), UStatus> {
+        // Get the data from UPayload
+        let Some(Data::Value(buf)) = payload.data else {
+            // TODO: Assume we only have Value here, no reference for shared memory
+            return Err(UStatus::fail_with_code(
+                UCode::INVALID_ARGUMENT,
+                "Invalid data",
+            ));
+        };
```
sophokles73: FMPOV it is not mandatory for a message to have a payload, is it?

src/lib.rs
```diff
@@ -0,0 +1,736 @@
...
```
sophokles73: FMPOV it is not mandatory for a message to have a payload, is it?
What about an RPC *turn_on_hvac* which simply returns with a successful comm_status?

src/lib.rs
```diff
@@ -0,0 +1,736 @@
...
+    }
+}
+
+#[async_trait]
+impl UTransport for UPClientZenoh {
+    async fn send(&self, message: UMessage) -> Result<(), UStatus> {
+        let payload = *message.payload.0.ok_or(UStatus::fail_with_code(
+            UCode::INVALID_ARGUMENT,
+            "Invalid uPayload",
+        ))?;
```
sophokles73: FMPOV it is not mandatory for a message to have a payload, is it?

src/lib.rs
evshary: The code is removed now. We decided to put the whole uAttributes directly into the user attachment.

sophokles73 (approved):
LGTM

stevenhartley (approved):
Great job @evshary thank you!

src/utransport.rs
```diff
@@ -0,0 +1,654 @@
...
+                    .validator()
+                    .validate(&attributes)
+                    .map_err(|e| {
+                        UStatus::fail_with_code(
+                            UCode::INVALID_ARGUMENT,
+                            format!("Wrong Publish UAttributes {e:?}"),
+                        )
+                    })?;
+                // Get Zenoh key
+                let topic = attributes.clone().sink;
```
PLeVasseur, line 485: Making a link to send() for publish style messages

src/utransport.rs
```diff
@@ -0,0 +1,654 @@
...
+                    .validator()
+                    .validate(&attributes)
+                    .map_err(|e| {
+                        UStatus::fail_with_code(
+                            UCode::INVALID_ARGUMENT,
+                            format!("Wrong Request UAttributes {e:?}"),
+                        )
+                    })?;
+                // Get Zenoh key
+                let topic = attributes.clone().sink;
```
PLeVasseur, line 501: Making a link to choice of sink/source UUri for request style messages

src/utransport.rs
```diff
@@ -0,0 +1,654 @@
...
+                    .validator()
+                    .validate(&attributes)
+                    .map_err(|e| {
+                        UStatus::fail_with_code(
+                            UCode::INVALID_ARGUMENT,
+                            format!("Wrong Response UAttributes {e:?}"),
+                        )
+                    })?;
+                // Get Zenoh key
+                let topic = attributes.clone().source;
```
PLeVasseur, line 517: Making a link to choice of sink/source UUri for response style messages

tests/integration_test.rs
```diff
@@ -0,0 +1,439 @@
...
+        Err(ustatus) => panic!("Internal Error: {:?}", ustatus),
+    };
+    let listener_string = upclient
+        .register_listener(uuri.clone(), Box::new(listener))
+        .await
+        .unwrap();
+
+    // Create uattributes
+    let mut attributes = UAttributesBuilder::publish(UPriority::UPRIORITY_CS4).build();
+    attributes.sink = Some(uuri.clone()).into();
```
PLeVasseur, line 209: Making a link to show there's a bit of a gap here -- the sink can be left blank for a publish style message, and if it were, currently a listener for this UUri could not find it, since the [sink is being used up above here](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/2/files#r1508078000) for publish-style message, which it should be the source.

I think you'd need to swap sink for source up above there @evshary

The [issue @sophokles73 opened up over here](https://github.com/eclipse-uprotocol/up-spec/issues/73) made me take a closer look.

src/rpc.rs
```diff
@@ -0,0 +1,137 @@
...
+            // TODO: Assume we only have Value here, no reference for shared memory
+            vec![]
+        };
+
+        // Generate UAttributes
+        // TODO: Check the ttl
+        let mut uattributes =
+            UAttributesBuilder::request(UPriority::UPRIORITY_CS4, topic.clone(), 255).build();
+        // TODO: How to create the source address for Response
+        let mut source = topic.clone();
```
PLeVasseur, line 59: Guess this was still TODO, so OK, to let thru for the moment. I think we talked about a week ago or so that on construction the UpClientZenoh instance would have to know enough about itself so that a source topic can be created. [Here's how it's done](https://github.com/eclipse-uprotocol/up-client-android-java/blob/c45f6306ba43a449c5cae39c9e2d0a2f62a4baa1/library/src/main/java/org/eclipse/uprotocol/UPClient.java#L204) over in `up-client-android-java`.

PR #5, opened 2024-03-01
Fix CI issues

evshary:
As titled, the PR is to fix the lint and build issue.
But we need to wait for the following PR to be merged.
https://github.com/eclipse-uprotocol/up-rust/pull/52

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,5 +20,5 @@ log = "0.4.17"
 prost = "0.12"
 prost-types = "0.12"
 protobuf = { version = "3.3" }
-up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", branch = "main" }
+up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", rev = "68c8a1d94f0006daf4ba135c9cbbfddcd793108d" }
 zenoh = { version = "0.10.1-rc", features = ["unstable"]}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,15 +15,11 @@ pub mod rpc;
 pub mod utransport;
 
 use protobuf::{Enum, Message};
-use std::io::Write;
 use std::{
     collections::HashMap,
     sync::{atomic::AtomicU64, Arc, Mutex},
 };
-use up_rust::{
-    uprotocol::{UAttributes, UCode, UMessage, UPayloadFormat, UPriority, UStatus, UUri},
-    uri::serializer::{MicroUriSerializer, UriSerializer},
-};
+use up_rust::uprotocol::{UAttributes, UCode, UMessage, UPayloadFormat, UPriority, UStatus, UUri};
 use zenoh::{
     config::Config,
     prelude::r#async::*,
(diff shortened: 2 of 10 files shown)


evshary:
Since [the PR](https://github.com/eclipse-uprotocol/up-rust/pull/52) is merged, this can be reviewed

Cargo.toml
```diff
@@ -20,5 +20,5 @@ log = "0.4.17"
 prost = "0.12"
 prost-types = "0.12"
 protobuf = { version = "3.3" }
-up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", branch = "main" }
+up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", rev = "68c8a1d94f0006daf4ba135c9cbbfddcd793108d" }
```
PLeVasseur, line 23: Ah okay, and then if there's a tag, we can use that instead. I'm always super impressed by Rust tooling. Using a git tag has been [supported since 2014](https://github.com/rust-lang/cargo/issues/126#issuecomment-48101852) :slightly_smiling_face:

src/lib.rs
```diff
@@ -70,39 +66,31 @@ impl UPClientZenoh {
...
-                buf.write_all(&id).map_err(|_| {
-                    UStatus::fail_with_code(UCode::INVALID_ARGUMENT, "Wrong authority")
-                })?;
-            } else if authority.has_ip() {
-                let ip = authority.ip().to_vec();
-                buf.write_all(&ip).map_err(|_| {
-                    UStatus::fail_with_code(UCode::INVALID_ARGUMENT, "Wrong authority")
-                })?;
-            }
+            let buf: Vec<u8> = authority.try_into().map_err(|_| {
```
PLeVasseur, line 71: :+1:

src/lib.rs
```diff
@@ -70,39 +66,31 @@ impl UPClientZenoh {
...
-            .fold(String::new(), |s, c| s + &format!("{c:02x}")))
     }
 
     // The UURI format should be "upr/<UAuthority id or ip>/<the rest of remote UUri>" or "upl/<local UUri>"
     fn to_zenoh_key_string(uri: &UUri) -> Result<String, UStatus> {
         if uri.authority.is_some() && uri.entity.is_none() && uri.resource.is_none() {
             Ok(String::from("upr/") + &UPClientZenoh::get_uauth_from_uuri(uri)? + "/**")
         } else {
-            let micro_uuri = MicroUriSerializer::serialize(uri).map_err(|_| {
+            let micro_uuri: Vec<u8> = uri.try_into().map_err(|_| {
```
PLeVasseur, line 93: :+1:

PLeVasseur (approved):
LGTM!

src/rpc.rs
```diff
@@ -52,16 +52,30 @@ impl RpcClient for UPClientZenoh {
         };
 
         // Generate UAttributes
-        // TODO: Check the ttl
-        let mut uattributes =
-            UAttributesBuilder::request(UPriority::UPRIORITY_CS4, topic.clone(), 255).build();
-        // TODO: How to create the source address for Response
+        let uuid_builder = UUIDBuilder::new();
```
PLeVasseur, line 55: I think this RPC flow still needs some thought though.

We should not be creating a new UUIDBuilder here and getting the reqid from it. I believe this is where the UUIDBuilder that's been constructed already for the uE sending the request should be used.

As is stated [over here in the UUID spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uuid.adoc#2-specification), the `rand_b` portion of the UUID is used to uniquely identify the uE, thus it should be "stable" and thus we should use the UUIDBuilder already associated with this uE.

I suppose that may mean that on construction, you'd need to construct a UUIDBuilder and then use it within `RpcClient::invoke_method()`.

WDYT?

Issue #6, opened 2024-03-01
Handle `Notification`-style `Publish` messages correctly in `send()`


PLeVasseur:
FYI @evshary -- I'm sensing we will need to modify `up-client-zenoh-rust` `UTransport::send()` to add being able to send  `Notification` with their Zenoh KeyExpr set at their sink instead of source. I'm referring specifically to [this part of the code](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/blob/cca6bdb761500b02717a480bb240b1f061cf3f29/src/utransport.rs#L485-L486).

`Notification`s are not an explicit message type, but are lumped in with `Publish`.

You can discriminate them based on whether there's a sink or not:
* No sink => it's actually `Publish`
* Has sink => it's actually `Notification`

(Does this seem safe to do @sophokles73? Any other way to discriminate between them?)

I thought of this when reading [Kai's comment](https://github.com/eclipse-uprotocol/up-spec/issues/73#issuecomment-1973461225).

PLeVasseur:
All done :slightly_smiling_face:

PR #9, opened 2024-03-05
Support Notification-style Publish.

evshary:
As @PLeVasseur mentioned in #6, support Notification-style Publish.

### DIFF
--- a/src/utransport.rs
+++ b/src/utransport.rs
@@ -482,7 +482,13 @@ impl UTransport for UPClientZenoh {
                         )
                     })?;
                 // Get Zenoh key
-                let topic = attributes.clone().source;
+                let topic = if attributes.sink.is_some() {
+                    // If sink is not empty, this is Notification => topic is sink
+                    attributes.clone().sink
+                } else {
+                    // If sink is empty, this is Publication => topic is source
+                    attributes.clone().source
+                };
                 let zenoh_key = UPClientZenoh::to_zenoh_key_string(&topic)?;
                 // Send Publish
                 self.send_publish(&zenoh_key, payload, attributes).await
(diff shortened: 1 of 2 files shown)


tests/integration_test.rs
```diff
@@ -26,24 +26,45 @@ use up_rust::{
 use uprotocol_zenoh::UPClientZenoh;
 use zenoh::config::Config;
 
-fn create_utransport_uuri() -> UUri {
-    UUri {
-        entity: Some(UEntity {
-            name: "body.access".to_string(),
-            version_major: Some(1),
-            id: Some(1234),
+fn create_utransport_uuri(index: u8) -> UUri {
```
PLeVasseur, line 29: Is the idea that you'd in the future keep adding more `UUri`s in here for testing purposes and you want to have a single factory function for creating them?

src/utransport.rs
```diff
@@ -482,7 +482,13 @@ impl UTransport for UPClientZenoh {
                         )
                     })?;
                 // Get Zenoh key
-                let topic = attributes.clone().source;
+                let topic = if attributes.sink.is_some() {
```
PLeVasseur, line 485: Yup, seems like this is what we need :slightly_smiling_face: 

Just wanna get some input from @sophokles73

sophokles73: It is correct to assume that you are dealing with a *Notification* if *sink* is not empty and a *Publish* otherwise.

PLeVasseur (approved):
Aside from the nit comment on the unit tests, I'll go ahead and approve :slightly_smiling_face:

tests/integration_test.rs
evshary: You're right. That's my intention.
Feel free to give me your opinions about this. I'm still wondering whether there is a more elegant way to do so.

Issue #10, opened 2024-03-06
Update RPC implementation to handle message `UUID`s correctly


PLeVasseur:
Highlighting [a comment I made on a previous PR](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/5/files#r1509121212):
> I think this RPC flow still needs some thought though.
> 
> We should not be creating a new UUIDBuilder here and getting the reqid from it. I believe this is where the UUIDBuilder that's been constructed already for the uE sending the request should be used.
> 
> As is stated [over here in the UUID spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uuid.adoc#2-specification), the `rand_b` portion of the UUID is used to uniquely identify the uE, thus it should be "stable" and thus we should use the UUIDBuilder already associated with this uE.
> 
> I suppose that may mean that on construction, you'd need to construct a UUIDBuilder and then use it within `RpcClient::invoke_method()`.
> 
> WDYT?

evshary:
Make sense. I'll see how to make the changes

evshary:
Close with #24

PR #12, opened 2024-03-07
UPClientZenoh should store the UUIDBuilder and reuse it.

evshary:
Solve #10 
UPClientZenoh should reuse its UUIDBuilder and keep the consistency of the UUID (lsb part).

### DIFF
--- a/examples/publisher.rs
+++ b/examples/publisher.rs
@@ -17,7 +17,6 @@ use up_client_zenoh::UPClientZenoh;
 use up_rust::{
     transport::{builder::UMessageBuilder, datamodel::UTransport},
     uprotocol::{UEntity, UPayloadFormat, UResource, UUri},
-    uuid::builder::UUIDBuilder,
 };
 use zenoh::config::Config;
 
@@ -55,7 +54,7 @@ async fn main() {
         let data = format!("{cnt}");
         let umessage = UMessageBuilder::publish(&uuri)
             .build_with_payload(
-                &UUIDBuilder::new(),
+                publisher.get_uuid_builder(),
                 data.as_bytes().to_vec().into(),
                 UPayloadFormat::UPAYLOAD_FORMAT_TEXT,
             )
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,7 +19,10 @@ use std::{
     collections::HashMap,
     sync::{atomic::AtomicU64, Arc, Mutex},
 };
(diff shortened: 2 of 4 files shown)


evshary:
@PLeVasseur I also reuse the UUIDBuilder while creating Publish messages.
Let me know your thoughts.
https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/12/files#diff-19929c8f4694b3b8b701d3bdfd81c3793f94a21737cde18e0d5e4ac39b878a80R57

PLeVasseur (approved):
Yup, I like your addition to a `UpClientZenoh` holding onto a `UUIDBuilder` for use by clients.

Also good that it's not used by default so that the uStreamer's usage will not be impacted :slightly_smiling_face:

stevenhartley (commented):
We made the UUIDBuilder a singleton inside of the language library, is there a reason why it is not done the same inside of Rust?

PLeVasseur:
> We made the UUIDBuilder a singleton inside of the language library, is there a reason why it is not done the same inside of Rust?

Seems possible, but it's going to be handled a bit differently in Rust due to the more restrictive / careful memory handling.

Seems like `lazy_static` or `once_cell` would be the way to go about it, from my reading.

For flavor:
```rust
use once_cell::sync::Lazy;
use std::sync::Mutex;

struct UUIDBuilder {
    // omitted
}

impl UUIDBuilder {
    fn new() -> UUIDBuilder {
        UUIDBuilder {
            // Initialize fields
        }
    }

    pub fn instance() -> &'static Mutex<UUIDBuilder> {
        static INSTANCE: Lazy<Mutex<UUIDBuilder>> = Lazy::new(|| {
            Mutex::new(UUIDBuilder::new())
        });
        &INSTANCE
    }
}
```

But probably should move this to an issue under `up-rust` and get feedback from others like @sophokles73

evshary:
Based on [the issue](https://github.com/eclipse-uprotocol/up-rust/issues/58), UUIDBuilder should be stored in up-rust.
Will create another PR while up-rust is ready.

PLeVasseur:
FYI @evshary -- I took a shot at implementing this here:
https://github.com/eclipse-uprotocol/up-rust/pull/74

And showed it can be integrated fairly easily with `up-client-zenoh-rust` here:
https://github.com/PLeVasseur/up-client-zenoh-rust/tree/feature/adjust_to_uuidbuilder_build_singleton

Feel free to review / check it out if you wanna :slightly_smiling_face:

evshary:
Thanks, I'll check it 😀

PR #16, opened 2024-03-12
Add LICENSE and CONTRIBUTING.

evshary:
Reminded by [the comments](https://github.com/eclipse-uprotocol/up-zenoh-example-rust/pull/1#pullrequestreview-1927637591), add the LICENSE and CONTRIBUTING for the project.


CONTRIBUTING.adoc
```diff
@@ -0,0 +1,36 @@
+= Contributing to Eclipse uProtocol
```
PLeVasseur, line 1: @sophokles73 reminded me we wanted to link to https://github.com/eclipse-uprotocol/.github/blob/main/CONTRIBUTING.adoc instead of re-writing this in each repo

evshary: Good point. I'll update it.

evshary: Done.

PLeVasseur (approved):
LGTM!

README.md
```diff
@@ -13,6 +13,10 @@ cargo build
 cargo test
 ```
 
+## Examples
+
+The examples of using up-client-zenoh-rust can be found in [up-zenoh-example-rust](https://github.com/eclipse-uprotocol/up-zenoh-example-rust).
```
sophokles73, line 18: is there a particular reason why you are putting the examples into a separate repository instead of just the `examples` folder (which seems to be common practice in Rust projects)?

evshary: I put the examples in the `examples` folder originally, but after the discussion in the weekly meeting, we decided to move the examples to another repository for every language.
Rust: https://github.com/eclipse-uprotocol/up-zenoh-example-rust
C++: https://github.com/eclipse-uprotocol/up-zenoh-example-cpp

The advantage is that users can know how to create a project more simply, but the disadvantage is that there will be many example repositories for every language and protocol.
I don't have strong opinions on the decision. Both are fine to me.

PR #17, opened 2024-03-12
Fix TTL issue

### LINKED ISSUE #14: ttl for RpcClient
Ensure that TTL from CallOptions are passed in the UAttributes and then the futures are completed exceptionally when the message times out (we did not get a reply)

evshary:
As title, the PR is to fix #14 issue


evshary:
@stevenhartley I think the PR can deal with #14

stevenhartley (approved):

PR #18, opened 2024-03-16
Use `up-rust` with `UListener` trait

PLeVasseur:
Hey @evshary :wave: 

This is definitely a WIP, but I will finish it.

I strongly want to test out the change I made to `up-rust` to align closer to the spec for uP-L1, which includes the `UListener` interface.

So far, so good with the parts I have modified and tested thus far:
* UTransport
* tests for UTransport

Still remaining to do:
- [ ] update rpc.rs
- [ ] update rpc tests
- [ ] get all integration tests working again
- [ ] point Cargo.toml for `up-rust` back to main repo if we think the approach taken in that PR is reasonable

Hope you don't feel I'm overstepping here, but I needed a way to verify that what I've done in `up-rust` will work in general for `up-client-foo-rust` implementations.

And hey -- as a side benefit `up-client-zenoh-rust` would be updated to agree with the latest in the `up-rust` repo :slightly_smiling_face:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,5 +20,8 @@ log = "0.4.17"
 prost = "0.12"
 prost-types = "0.12"
 protobuf = { version = "3.3" }
-up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", rev = "68c8a1d94f0006daf4ba135c9cbbfddcd793108d" }
+# upstream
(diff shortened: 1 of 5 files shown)


evshary:
That's fine with me. 🙂

PLeVasseur:
Closing in favor of #19, which has the _instance-specific_ behavior around UListener that we want.

PR #20, opened 2024-03-19
Initialize UPClientZenoh with different configurations

evshary:
As discussed in https://github.com/eclipse-uprotocol/up-spec/issues/85
we should provide a way for users to add configuration into `UPClientZenoh`.

Now we have three ways:
* `UPClientZenoh::new()`: Using default configuration
* `UPClientZenoh::new_with_config(config: UPClientZenohConfig)`: Using UPClientZenohConfig, which is simpler and limited
* `UPClientZenoh::new_with_zenoh_config(config: Config)`: Using Zenoh config, which is mainly for advanced users

I created the PR to discuss whether this is what we want.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -32,6 +32,25 @@ pub type UtransportListener = Box<dyn Fn(Result<UMessage, UStatus>) + Send + Syn
 
 const UATTRIBUTE_VERSION: u8 = 1;
 
+pub enum ZenohMode {
+    Peer,
+    Client,
+    Router,
+}
+
+/// Configuration struct for `UPClientZenoh`
+pub struct UPClientZenohConfig {
+    /// The Zenoh's mode (router, peer or client)
+    pub mode: ZenohMode,
+    /// Which endpoints to listen on. E.g. tcp/localhost:7447.
+    /// By configuring the endpoints, it is possible to tell Zenoh which are the endpoints that other routers,
+    /// peers, or client can use to establish a zenoh session.
+    pub listen: Option<Vec<String>>,
+    /// Which endpoints to connect to. E.g. tcp/localhost:7447.
+    /// By configuring the endpoints, it is possible to tell Zenoh which router/peer to connect to at startup.
+    pub connect: Option<Vec<String>>,
+}
+
 pub struct ZenohListener {}
 pub struct UPClientZenoh {
     session: Arc<Session>,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -32,6 +32,25 @@ pub type UtransportListener = Box<dyn Fn(Result<UMessage, UStatus>) + Send + Syn
...
+    Peer,
+    Client,
+    Router,
+}
+
+/// Configuration struct for `UPClientZenoh`
+pub struct UPClientZenohConfig {
+    /// The Zenoh's mode (router, peer or client)
+    pub mode: ZenohMode,
+    /// Which endpoints to listen on. E.g. tcp/localhost:7447.
```
PLeVasseur, line 45: My big question is "how clueless should the average uE developer be allowed to be?" :slightly_smiling_face: 

Should we write another API so that it "just works" as expected for 99% of cases where someone is writing a typical uE?

By that I mean -- if someone is writing a "typical" uE using `up-client-zenoh-rust`, they most likely are fine with having a default Zenoh Config, aside from having their mode set to Peer, right?

All of these boxed in pink:
image

For those, perhaps a simpler API would be a little nicer, such as:
```rust
impl UPClientZenoh {
    pub fn new_peer() {
        let mut zenoh_config = Config::default();
        zenoh_config
            .set_mode(WhatAmI::Peer);
        UPClientZenoh::new_with_zenoh_config(zenoh_config).await
    }
}
```

Then we'd have:
1. Most simple API (`new_peer()`), just for making a typical Peer, that would be used by 99% of uE developers
2. Slightly more advanced API with enough flexibility for 99% of the remaining 1% (`new_with_config()`)
3. Most advanced API that gives access to all configurability of Zenoh (`new_with_zenoh_config`) for the remaining 0.01% of use cases

What do you think?

evshary: In fact, the default config is using `Peer` mode and that's why I don't configure it explicitly. 😃

PLeVasseur: I bow down before your API design haha 😆

PLeVasseur:
Based on today's discussion, speaking from a Rust perspective (I know something similar is possible in C++), how about something like what @tamarafischer & @evshary were proposing for a stable interface:
```rust
struct UPClientZenoh { 
    // ...
}

impl UPClientZenoh {
    pub fn new() -> Self {
        // ...
    }

    pub fn new_with_config(config: Config) -> Self {
        // ...
    }
}
```

And then considering the two requirements I heard from @stevenhartley:
1. There should be specialized constructors for uEs in certain parts of the reference topology
2. In production we should only allow using those specialized constructors
how about something like this @evshary?

```
pub struct UPZenohClient {
    up_client_zenoh: Arc<UPClientZenoh>
}

impl UPZenohClient {
    pub async fn new(connect_endpoint: Endpoint) -> Result<Self, UStatus> {
        let mut config = Config::default();
        /* use connect_endpoint here to set proper field in config */
        let up_client_zenoh = UPClientZenoh::new(config).await?;
        Ok(Self {
            up_client_zenoh: Arc::new(up_client_zenoh)
        })
    }
}

impl Deref for UPZenohClient {
    type Target = Arc<UPClientZenoh>;

    fn deref(&self) -> &Self::Target {
        &self.up_client_zenoh
    }
}
```

And in end-user code when writing a uE their life simplifies to:
```rust
// bespoke configuration stuff hidden away
let my_cool_peer = UPZenohClient::new(my_connection_endpoint);
// ...

// able to use the same APIs as UPZenohRust transparently thanks to impl'ing Deref
let result = my_cool_peer.register_listener(uuri, callback);
my_cool_peer.send(message);
```

What does this get us?
1. Allows there to be a "core" `UPClientZenoh` that will not see _any_ churn as requirements change regarding topology & the specific configurations for different kinds of uEs
2. Each new `UPZenohFooRole` is very simple and involves only adding the 20 or so lines of code for each type. In reality we're bound to only have a few typical types in the topology
3. If an end-user would like to, they could even perform this same pattern for other types if they had a `UPZenohMyVerySpecificNicheCase`
4. In production we can lint and deny direct usage of `UPClientZenoh` and only allow those `UPZenohFooRole` that are deemed safe for end-users
5. If we'd like to, `UPZenohFooRole`s don't even have to live within the `up-client-zenoh-rust` repo. We could put the few that we think we'd need for this first topology, if we want to, though.

evshary:
Let's close it now and collect more information from users in the future.

PR #21, opened 2024-03-26
Update up-client-zenoh-rust with up-spec 1.5.7

evshary:
Since up-rust is upgraded to up-spec 1.5.7 in https://github.com/eclipse-uprotocol/up-rust/pull/66, the PR is to keep the code updated.
I'll update https://github.com/eclipse-uprotocol/up-zenoh-example-rust later


evshary:
@PLeVasseur I think the PR is ready. There are a lot of changes, but it should not affect your uStreamer implementation since I didn't change the API (except removing RpcServer)

Here are the modifications:
* Upgrade to the latest up-rust to match up-spec 1.5.7
* Update the way of error logging
* Support initializing the constructor with UUri (will generate randomly if we don't assign it)
* Reorganize the test code

Hope it's compatible with your uStreamer implementation

src/lib.rs
```diff
@@ -43,18 +46,83 @@ pub struct UPClientZenoh {
...
+    /// # Examples
+    ///
+    /// ```
+    /// # async_std::task::block_on(async {
+    ///     use up_client_zenoh::UPClientZenoh;
+    ///     use zenoh::config::Config;
+    ///     let upclient = UPClientZenoh::new(Config::default()).await.unwrap();
+    /// # });
+    /// ```
     pub async fn new(config: Config) -> Result<UPClientZenoh, UStatus> {
```
PLeVasseur, line 75: Hmmm, I dunno.

Feels like this could be a bad idea to open up to end-users since they may "do the wrong thing" and use this and get confused when `RpcClient::invoke_method()` isn't working like they expected.

Would rather not expose any constructor publicly that could cause this issue.

WDYT?

src/lib.rs
```diff
@@ -43,18 +46,83 @@ pub struct UPClientZenoh {
...
+    ///             version_minor: None,
+    ///             ..Default::default()
+    ///         })
+    ///         .into(),
+    ///         ..Default::default()
+    ///     };
+    ///     let upclient = UPClientZenoh::new_with_uuri(Config::default(), uuri).await.unwrap();
+    /// # });
+    /// ```
+    pub async fn new_with_uuri(config: Config, uuri: UUri) -> Result<UPClientZenoh, UStatus> {
```
PLeVasseur, line 121: Hmm, again I'm thinking of the 99% case of people using this library. They're not gonna want to have to think about what UUri to put or remember that they're supposed to put "rpc.response" in the `UResource`.

IMHO this should probably explicitly take `UAuthority`, `UEntity`, and the `version_major` to do the right thing for the end user.

PLeVasseur: On the other hand, I like that you _store_ it as a `UUri`, that makes sense :slightly_smiling_face:

src/lib.rs
```diff
@@ -43,18 +46,83 @@ pub struct UPClientZenoh {
     query_map: Arc<Mutex<HashMap<String, Query>>>,
     // Save the callback for RPC response
     rpc_callback_map: Arc<Mutex<HashMap<String, Arc<UtransportListener>>>>,
+    // Used to identify different callback
     callback_counter: AtomicU64,
+    // Source UUri in RPC
+    uuri: UUri,
```
PLeVasseur, line 52: IMHO let's name it `rpc_request_source_uuri` or something similarly descriptive, rather than that be in the comment.

WDYT?

src/lib.rs
```diff
@@ -43,18 +46,83 @@ pub struct UPClientZenoh {
...
+    /// # });
+    /// ```
+    pub async fn new_with_uuri(config: Config, uuri: UUri) -> Result<UPClientZenoh, UStatus> {
         let Ok(session) = zenoh::open(config).res().await else {
-            return Err(UStatus::fail_with_code(
-                UCode::INTERNAL,
-                "Unable to open Zenoh session",
-            ));
+            let msg = "Unable to open Zenoh session".to_string();
+            log::error!("{msg}");
```
PLeVasseur, line 124: I like the added logging :+1:

src/lib.rs
```diff
@@ -148,30 +236,27 @@ impl UPClientZenoh {
     fn attachment_to_uattributes(attachment: &Attachment) -> anyhow::Result<UAttributes> {
         let mut attachment_iter = attachment.iter();
         if let Some((_, value)) = attachment_iter.next() {
-            let version = *value.as_slice().first().ok_or(UStatus::fail_with_code(
-                UCode::INTERNAL,
-                "uAttributes version is empty",
-            ))?;
+            let version = *value.as_slice().first().ok_or_else(|| {
+                let msg = "UAttributes version is empty".to_string();
```
PLeVasseur, line 240: WDYT about adding some identifying information as to which message had these failures, e.g. logging at a minimum the `UMessage.attributes.id`?

I think it would help traceability for an end-user

src/rpc.rs
```diff
@@ -32,91 +29,77 @@ impl RpcClient for UPClientZenoh {
...
-                "Unable to create uAttributes",
-            )));
-        };
+        let mut uattributes = UAttributes::request(
+            uuid_builder.build(),
+            topic,
+            self.get_response_uuri(),
+            options.clone(),
+        );
+        // TODO: We should not have reqid here, but validate Request will fail
```
PLeVasseur, line 53: Does this mean you need to open a PR on `up-rust` to fix this?

src/rpc.rs
```diff
@@ -32,91 +29,77 @@ impl RpcClient for UPClientZenoh {
...
-        let Ok(UMessage {
-            attributes: uattributes,
-            ..
-        }) = umessage
-        else {
-            return Err(RpcMapperError::UnexpectedError(String::from(
-                "Unable to create uAttributes",
-            )));
-        };
+        let mut uattributes = UAttributes::request(
```
PLeVasseur, line 47: Using the function @sophokles73 added :rocket: :rocket: :rocket: :rocket: :rocket:

src/lib.rs
evshary: Good idea. I'll add it

src/utransport.rs
```diff
@@ -514,33 +505,32 @@ impl UTransport for UPClientZenoh {
...
-                        UStatus::fail_with_code(
-                            UCode::INVALID_ARGUMENT,
-                            format!("Wrong Response UAttributes {e:?}"),
-                        )
+                        let msg = format!("Wrong Response UAttributes: {e:?}");
+                        log::error!("{msg}");
+                        UStatus::fail_with_code(UCode::INVALID_ARGUMENT, msg)
                     })?;
                 // Get Zenoh key
                 let topic = attributes.clone().source;
```
PLeVasseur, line 517: When I was reading the [`up-spec` changes](https://github.com/eclipse-uprotocol/up-spec/blob/c0ea3bfb994a0e4db46dafca0e7b9d6bd7ba2b3b/up-l1/zenoh.adoc#254-transfer-response-message) you had put together, in this section on Response messages, you mention you should use `sink`.

Shouldn't this be `sink` then?

src/utransport.rs
```diff
@@ -593,18 +583,20 @@ impl UTransport for UPClientZenoh {
             // This is special UUri which means we need to unregister both listeners
```
PLeVasseur, line 583:
```rust
// This is special UUri which means we need to unregister both listeners
```
=>
```rust
// This is special UUri which means we need to unregister all listeners
```
?

src/lib.rs
evshary: You're right. Although generating UUri automatically is easy for users to use (to whom they know what is going on), it'll confuse the newcomer. Okay, I'll remove it.

src/utransport.rs
```diff
@@ -514,33 +505,32 @@ impl UTransport for UPClientZenoh {
...
-        Err(UStatus::fail_with_code(
-            UCode::UNIMPLEMENTED,
-            "Not implemented",
-        ))
+        let msg = "Not implemented".to_string();
+        log::error!("{msg}");
+        Err(UStatus::fail_with_code(UCode::UNIMPLEMENTED, msg))
     }
 
     async fn register_listener(
```
PLeVasseur, line 536: btw -- do we need to have now a `register_notification_listener()`? Or is that handled by one of the existing functions?

src/lib.rs
PLeVasseur, line 121: WDYT about trying to initialize the `env_logger` upon construction and just quietly failing if it's already been initialized?

I saw some Zenoh plugin code do this and made sense to me.

tests/publish.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    // Initialization
+    let target_data = String::from("Hello World!");
+    let upclient = UPClientZenoh::new(Config::default()).await.unwrap();
+    let uuri = test_lib::create_utransport_uuri(0);
+
+    // Register the listener
+    let uuri_cloned = uuri.clone();
+    let data_cloned = target_data.clone();
+    let listener = move |result: Result<UMessage, UStatus>| match result {
```
PLeVasseur, line 34: Hmmm. One thing I'm not sure about with doing the testing this way.

Doesn't this mean that if the message never makes it to the `listener` function, then the test will pass?

i.e. if we give a malformed `UUri` into the `UMessage` we `send()` below, then the test will just pass, right?

tests/publish.rs
```diff
@@ -0,0 +1,120 @@
...
+        }
+        Err(ustatus) => panic!("Internal Error: {ustatus:?}"),
+    };
+    let listener_string = upclient
+        .register_listener(dst_uuri.clone(), Box::new(listener))
+        .await
+        .unwrap();
+
+    // Send UMessage
+    // TODO: up_rust has bugs while creating UMessageBuilder::notification
```
PLeVasseur, line 102: PR into `up_rust` possible? :slightly_smiling_face:

src/lib.rs
PLeVasseur: In that case it'd be a different `UUri` tho because for an RPC response, the `UResource` should have "rpc.response" [according to the spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#3-validators):
image
whereas the Notification would not

PLeVasseur:
OK @evshary -- done with my first pass :+1:

evshary:
> OK @evshary -- done with my first pass 👍

Thank you. I'll check your comments.

src/lib.rs
evshary: Yes, I mean we can store the UUri in UPClientZenoh and provide two functions to get the UUri

* `get_response_uuri`: Will get the UUri and replace the UResource with `rpc.response`
* `get_uuri`: Will get the original UUri

When calling `invoke_method`, we should fill the source address with `get_response_uuri`
When calling `send(notification)`, we should fill the source address with `get_uuri`

Do you think it makes sense?

PLeVasseur: > * `get_response_uuri`: Will get the UUri and replace the UResource with `rpc.response`

> When calling `invoke_method`, we should fill the source address with `get_response_uuri`

Sure -- this would happen within the library itself (within `RpcClient::invoke_method()`), no exposure to the end-user, right? So this function could be `pub (crate)`?

> * `get_uuri`: Will get the original UUri

> When calling `send(notification)`, we should fill the source address with `get_uuri`

Under-the-hood, within `send()`?

Or the end-user would call `get_uuri()` and use that in constructing their Notification message?

I'd lean toward the latter since nothing in the spec says we should be rewriting that in a `up-client-foo-bar` when calling `UTransport::send()`.

-------

In any case -- I think the API should still explicitly ask for `UAuthority`, `UEntity`, and `version_major` to make explicit what we're doing.

WDYT?

evshary: IMHO, I think `env_logger` is suitable to be used in binary. The library users (no matter the tests, examples, etc.) can decide whether they want to init the logger or how they init the logger by themselves.

src/utransport.rs
evshary: Thank you! The legacy comments I didn't notice before. 😆

src/lib.rs
evshary: > Sure -- this would happen within the library itself (within RpcClient::invoke_method()), no exposure to the end-user, right? So this function could be pub (crate)?

Not only. We need to use this if we want to call `send(request)` instead of `invoke_method()`
For example:
https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/21/files#diff-b2d3f8f8c9e7487b7199b4393c1f706d07023da1bf6ca24d667b0096e88422d2R125

> I'd lean toward the latter since nothing in the spec says we should be rewriting that in a up-client-foo-bar when calling UTransport::send().

Yes, I mean the latter.

> In any case -- I think the API should still explicitly ask for UAuthority, UEntity, and version_major to make explicit what we're doing.

I'm not sure. Do you think users will call `get_uuri()` to construct the notification message? (I believe notification message should include resource) If not, then I agree with you should only have UAuthority and UEntity.

BTW, why do we need to have `version_major` separately? Shouldn't it be the member of `UEntity`?

src/utransport.rs
evshary: No. I think we have no way to distinguish publish and notification only from UUri.
We can only handle them with `register_publish_listener` together. (Maybe need a better name?)

tests/publish.rs
evshary: You're right. Let me see whether there is a better way to handle it.

src/utransport.rs
evshary: It seems like there is a little mismatch. Let me figure it out.

src/lib.rs
PLeVasseur: > Not only. We need to use this if we want to call send(request) instead of invoke_method()

Ah, you're right! I missed this case :slightly_smiling_face:

> Yes, I mean the latter.

:+1: 

> I'm not sure. Do you think users will call get_uuri() to construct the notification message? (I believe notification message should include resource) If not, then I agree with you should only have UAuthority and UEntity.

Hmm. I'm not too sure frankly, it's early and we have no users yet :smile: 

But I wanted to highlight that there are a few aspects to this:
1. What should the API be for `new()`?
   * I think it should take UAuthority and UEntity because those are the only stable elements between the Request source UUri and the Notification source UUri
2. How should people be forming the source UUri in a UMessage for Request?
   * I really like the example you showed for this. Every portion of the UUri is fixed, so this makes a lot of sense.
3. How should people be forming UMessage for Notification?
   * The UAuthority and UEntity are fixed, the UResource is not.
   * The "intention-revealing" option almost seems to me to be to have a function like `get_notification_src_uuri(resource: UResource)` which will return a fully-formed UUri with the UAuthority, UEntity, and UResource

WDYT?

> BTW, why do we need to have `version_major` separately? Shouldn't it be the member of `UEntity`?

Ignore, sorry, I was confused haha.

src/utransport.rs
PLeVasseur: Okay. I think you're right, some name change may be needed.

tests/publish.rs
evshary: Should be fixed [here](https://github.com/eclipse-uprotocol/up-rust/pull/75)

tests/publish.rs
evshary: Done. I found that the notification always failed and I didn't notice that before 😂

src/utransport.rs
evshary: Done

src/lib.rs
evshary: > it's early and we have no users yet

I don't have a strong opinion on using UUri or UAuthority+UEntity. I'll use the latter and we can see if users like it or not in the future.

src/lib.rs
evshary: Since the uuri is used in both notification and RPC request, maybe the name should be something like `source_uuri`
Naming is always difficult for me 😅

PLeVasseur: Yup, makes sense now that I understand it's going to be used for both Notification and Request source UUris :slightly_smiling_face:

src/lib.rs
evshary: Done

src/lib.rs
evshary: Done.

src/lib.rs
evshary: Done.

src/utransport.rs
evshary: Update the spec and code for alignment
https://github.com/eclipse-uprotocol/up-spec/blob/a3681e3a9f3bb824b75315f276281088c0c8e605/up-l1/zenoh.adoc#254-transfer-response-message

evshary:
@PLeVasseur I think the PR is ready for your second review :smile:

tests/publish.rs
```diff
@@ -61,6 +63,9 @@ async fn test_publish_and_subscribe() {
     // Waiting for the subscriber to receive data
     task::sleep(time::Duration::from_millis(1000)).await;
 
+    // Compare the result
+    assert_eq!(*verified_data.lock().unwrap(), target_data);
```
PLeVasseur, line 65: Yup, makes sense. Good way to ensure we're passing the test :slightly_smiling_face:

tests/publish.rs
```diff
@@ -117,8 +116,9 @@ async fn test_notification_and_subscribe() {
     task::sleep(time::Duration::from_millis(1000)).await;
 
     // Compare the result
-    // TODO: Need to fix up_rust first
-    //assert_eq!(*verified_data.lock().unwrap(), target_data);
+    // TODO: Wait for up-rust to update
+    // https://github.com/eclipse-uprotocol/up-rust/pull/75
+    assert_eq!(*verified_data.lock().unwrap(), target_data);
```
PLeVasseur: :+1:

src/lib.rs
```diff
@@ -49,7 +49,7 @@ pub struct UPClientZenoh {
     // Used to identify different callback
     callback_counter: AtomicU64,
     // Source UUri in RPC
-    uuri: UUri,
+    source_uuri: UUri,
```
PLeVasseur, line 52: :+1:

src/lib.rs
```diff
@@ -102,31 +103,38 @@ impl UPClientZenoh {
...
     ///         ..Default::default()
     ///     };
-    ///     let upclient = UPClientZenoh::new_with_uuri(Config::default(), uuri).await.unwrap();
+    ///     let upclient = UPClientZenoh::new_with_uuri(Config::default(), uauthority, uentity).await.unwrap();
     /// # });
     /// ```
     pub async fn new_with_uuri(
         config: Config,
-        source_uuri: UUri,
+        uauthority: UAuthority,
```
PLeVasseur, line 91: :+1:

PLeVasseur, line 91: I'm wondering if we should enforce that the `UAuthority` and `UEntity` passed in pass the Micro Form tests and return a failure otherwise? WDYT?

PLeVasseur:
> @PLeVasseur I think the PR is ready for your second review 😄

Finished round 2 -- looking good :slightly_smiling_face: 

Had one item that I think would be good to check on:
https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/21#discussion_r1548087422

src/lib.rs
evshary: Good catch! I added the validator and also use `testcase` to add some testcases, which I learned from up-rust :smile:

evshary:
@PLeVasseur Thank you! Let me know if you have any other comments.

src/utransport.rs
```diff
@@ -441,71 +420,77 @@ impl UPClientZenoh {
...
-                    attributes.clone().sink
-                } else {
-                    // If sink is empty, this is Publication => topic is source
-                    attributes.clone().source
-                };
+                // Get Zenoh key: Publication => topic is source
+                let topic = attributes.clone().source;
+                let zenoh_key = UPClientZenoh::to_zenoh_key_string(&topic)?;
+                // Send Publish
+                self.send_publish(&zenoh_key, payload, attributes).await
```
PLeVasseur, line 467: `self.send_publish()` => `self.send_publish_notification()`?

src/utransport.rs
```diff
@@ -441,71 +420,77 @@ impl UPClientZenoh {
...
+                //    .map_err(|e| {
+                //        let msg = format!("Wrong Notification UAttributes: {e:?}");
+                //        log::error!("{msg}");
+                //        UStatus::fail_with_code(UCode::INVALID_ARGUMENT, msg)
+                //    })?;
+                // Get Zenoh key: Notification => topic is sink
+                let topic = attributes.clone().sink;
                 let zenoh_key = UPClientZenoh::to_zenoh_key_string(&topic)?;
                 // Send Publish
                 self.send_publish(&zenoh_key, payload, attributes).await
```
PLeVasseur, line 484: Same comment as [here](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/21/files#r1549478752)

PLeVasseur:
> @PLeVasseur Thank you! Let me know if you have any other comments.

Hey @evshary -- I missed [this tiny nitpick](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/21#discussion_r1549478752) which I think gives symmetry. Could you take a look?

src/utransport.rs
evshary: Done

src/utransport.rs
evshary: Done.

evshary:
@PLeVasseur Done and your comment also reminds me to remove some TODO in code since https://github.com/eclipse-uprotocol/up-rust/pull/75 is merged

PLeVasseur (approved):
LGTM

evshary:
@stevenhartley Could you help me to  merge the PR? I would like to work on https://github.com/eclipse-uprotocol/up-spec/pull/103 based on the PR's changes.

Issue #22, opened 2024-04-08
Update to use `UListener` interface change within `UTransport`


PLeVasseur:
Hey @evshary -- I think you were saying you had a feature branch updating to use `UListener`.

Could you open a PR supporting those changes now that they have merged?
* https://github.com/eclipse-uprotocol/up-rust/pull/69
* https://github.com/eclipse-uprotocol/up-rust/pull/79
* https://github.com/eclipse-uprotocol/up-rust/pull/81

evshary:
I've heard from @AnotherDaniel that he needs this. I'll work on this.

evshary:
Close with #24

PR #24, opened 2024-04-10
Update UListener interface and UUID builder

evshary:
The PR is to update up-rust to the latest version, including the following changes:

* Use the new UListener interface: #22
* Use the singleton UUID builder in up-rust: #10

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,11 +17,11 @@ pub mod utransport;
 use protobuf::{Enum, Message};
 use std::{
     collections::HashMap,
-    sync::{atomic::AtomicU64, Arc, Mutex},
+    sync::{Arc, Mutex},
 };
 use up_rust::{
-    UAttributes, UAuthority, UCode, UEntity, UMessage, UPayloadFormat, UPriority, UResourceBuilder,
-    UStatus, UUri,
+    ComparableListener, UAttributes, UAuthority, UCode, UEntity, UListener, UPayloadFormat,
+    UPriority, UResourceBuilder, UStatus, UUri,
 };
 use zenoh::{
     config::Config,
@@ -31,23 +31,22 @@ use zenoh::{
     subscriber::Subscriber,
 };
 
-pub type UtransportListener = Box<dyn Fn(Result<UMessage, UStatus>) + Send + Sync + 'static>;
-
 const UATTRIBUTE_VERSION: u8 = 1;
 
-pub struct ZenohListener {}
+type SubscriberMap = Arc<Mutex<HashMap<(UUri, ComparableListener), Subscriber<'static, ()>>>>;
+type QueryableMap = Arc<Mutex<HashMap<(UUri, ComparableListener), Queryable<'static, ()>>>>;
+type QueryMap = Arc<Mutex<HashMap<String, Query>>>;
+type RpcCallbackMap = Arc<Mutex<HashMap<UUri, Arc<dyn UListener>>>>;
 pub struct UPClientZenoh {
     session: Arc<Session>,
     // Able to unregister Subscriber
-    subscriber_map: Arc<Mutex<HashMap<String, Subscriber<'static, ()>>>>,
+    subscriber_map: SubscriberMap,
     // Able to unregister Queryable
-    queryable_map: Arc<Mutex<HashMap<String, Queryable<'static, ()>>>>,
+    queryable_map: QueryableMap,
     // Save the reqid to be able to send back response
-    query_map: Arc<Mutex<HashMap<String, Query>>>,
+    query_map: QueryMap,
     // Save the callback for RPC response
-    rpc_callback_map: Arc<Mutex<HashMap<String, Arc<UtransportListener>>>>,
-    // Used to identify different callback
-    callback_counter: AtomicU64,
+    rpc_callback_map: RpcCallbackMap,
     // Source UUri in RPC
     source_uuri: UUri,
 }
@@ -117,7 +116,6 @@ impl UPClientZenoh {
             queryable_map: Arc::new(Mutex::new(HashMap::new())),
             query_map: Arc::new(Mutex::new(HashMap::new())),
             rpc_callback_map: Arc::new(Mutex::new(HashMap::new())),
-            callback_counter: AtomicU64::new(0),
             source_uuri,
         })
     }
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -44,7 +44,7 @@ impl RpcClient for UPClientZenoh {
 
         // Create UAttributes and put into Zenoh user attachment
         let uattributes = UAttributes::request(
-            UUIDBuilder::new().build(),
+            UUIDBuilder::build(),
             topic,
             self.get_response_uuri(),
             options.clone(),
--- a/src/utransport.rs
+++ b/src/utransport.rs
@@ -11,15 +11,13 @@
 // Contributors:
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
-use crate::{UPClientZenoh, UtransportListener};
+use crate::UPClientZenoh;
+use async_std::task::block_on;
 use async_trait::async_trait;
-use std::{
-    sync::{atomic::Ordering, Arc},
-    time::Duration,
-};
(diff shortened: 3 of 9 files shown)


evshary:
@AnotherDaniel and @PLeVasseur I'm still fixing the tests in the PR, but you can take a look if you need this.

evshary:
@AnotherDaniel and @PLeVasseur You can review the PR now. Since Daniel wants this urgently, I'll deal with some minor issues in another PR.

src/utransport.rs
```diff
@@ -102,65 +100,71 @@ impl UPClientZenoh {
...
                     log::error!("{msg}");
-                    Err(UStatus::fail_with_code(UCode::INTERNAL, msg))
+                    block_on(resp_callback.on_error(UStatus::fail_with_code(UCode::INTERNAL, msg)));
                 }
-            };
-            resp_callback(msg);
+            }
         };
 
         // Send query
```
PLeVasseur, line 170: I'm not great at Zenoh :)

I'm trying to understand if there's any issue with us using the `zenoh_key` here alone. Is it possible that if we fired off two requests, one after the other that we'd end up getting them back in the wrong order? Should we make use of the `req_id` at all?

Sorry for newb questions :upside_down_face:

src/rpc.rs
```diff
@@ -44,7 +44,7 @@ impl RpcClient for UPClientZenoh {
 
         // Create UAttributes and put into Zenoh user attachment
         let uattributes = UAttributes::request(
-            UUIDBuilder::new().build(),
+            UUIDBuilder::build(),
             topic,
             self.get_response_uuri(),
             options.clone(),
```
PLeVasseur, line 50: Sorry, this part is not in the section I can leave a review comment on, but again something I noticed and wanted to check on.

Same kind of question as in [this comment](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/24/files?diff=split&w=0#r1563253578)

PLeVasseur, line 50: Or perhaps... slightly different question, if we call `invoke_method()` once, are we able to call `invoke_method()` once again immediately thereafter, or would this line of code:
```rust
        let getbuilder = self
            .session
```
block the second invocation until the first had returned?

Sorry, again, I'm a newb to Zenoh :)

src/utransport.rs
evshary: Zenoh will take care of the order of queries and not mix them. That's why we don't need to use `req_id` while sending the query.

src/rpc.rs
evshary: Same as [here](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/24#discussion_r1565149160). `req_id` is only used in the uProtocol layer and Zenoh doesn't need it in the transmission.

evshary: If we want to call `invoke_method` twice at the same time, I think we can use something like `async_std::task::spawn`. Then the two `invoke_method` will not interfere with each other.

If we have something like `invoke_method(...).await`, the second one still needs to wait for the first one.

src/utransport.rs
PLeVasseur: Could you explain a little detail behind how the queries are ordered correctly?

If we sent out two Request messages one immediately after the other, call them A and B. And a Response messages was sent back for B first, how does the ordering get enforced?

If there's some bit of documentation or better yet code / tests that show this I think that'd be great.

src/rpc.rs
PLeVasseur: Thx. Makes sense

src/utransport.rs
evshary: Each time we create a Zenoh query to send out, there is also a `qid` inside Zenoh, which is used to differentiate different messages. While receiving the response message, Zenoh will use the `qid` to find the corresponding buffer to put the message.

In your case, we called `invoke_method(A)` and `invoke_method(B)`. `invoke_method(A)` should only get the response message from its buffer with the correct `qid`. Response message to B will never go inside the buffer of A.

Since this is the mechanism inside Zenoh, I prefer to hide it instead of exposing to users.

PLeVasseur: Ahhh okay. So even though the `zenoh_key` may be the same, there's the `qid` in the implementation of Zenoh that will correctly tie together each request with the corresponding response. Makes sense.

---------

Maybe not on this PR, but I think it'd be nice to show executing a couple of `invoke_method()` one after the other and showing they come back in the correct order. Ensuring in the test that we `invoke_method()` on another task so we can immediately call it again and show that even though we fired request_A, _request_B and were returned response_B, response_A they were correlated correctly.

src/lib.rs
```diff
@@ -17,11 +17,11 @@ pub mod utransport;
 use protobuf::{Enum, Message};
 use std::{
     collections::HashMap,
-    sync::{atomic::AtomicU64, Arc, Mutex},
+    sync::{Arc, Mutex},
```
PLeVasseur, line 20: When I've been developing async code I've starting using the `Mutex` implementation provided by the async framework because then the `lock()` is async aware so it's possible to hold across an `.await` point.

Maybe use `async_std::sync::Mutex`? What are your thoughts?

PLeVasseur:
Hey @evshary -- was [this](https://github.com/eclipse-uprotocol/up-streamer-rust/pull/8#discussion_r1560379306) addressed in this PR? Will be done in another?

Could you point me to where you'd need to update?

tests/publish.rs
```diff
@@ -14,111 +14,127 @@
...
-                panic!("The message should be Data::Value type.");
-            }
-        }
-        Err(ustatus) => panic!("Internal Error: {ustatus:?}"),
-    };
-    let listener_string = upclient
-        .register_listener(uuri.clone(), Box::new(listener))
+    let pub_listener = Arc::new(PublishNotificationListener::new());
+    upclient_recv
+        .register_listener(dst_uuri.clone(), pub_listener.clone())
```
PLeVasseur, line 70:
```suggestion
        .register_listener(origin_uuri.clone(), pub_listener.clone())
```
?

src/utransport.rs
evshary: It's a good suggestion, but I'm not sure how to make the response always in reverse order in the test.
Do you have any thoughts?

tests/publish.rs
evshary: Yes, this one. Thank you for pointing out. I think I should rename `dst_uuri` to `listen_uuri` for better understanding.

src/lib.rs
evshary: If we want to hold the `Mutex` across a `.await` point, we should use `async_std::sync::Mutex`.
But in my case, I don't need to hold the Mutex. I just want to make sure the mutual exclusivity of `HashMap` and no matter insert / remove can be done without holding the Mutex across `.await` point.

I can still use `async_std::sync::Mutex` in this case, but then I need to make sure the Mutex is always used in the async function, which increase the complexity.

PLeVasseur: Ah okay. Interesting trade-off. Thanks for the reasoning. :+1:

src/utransport.rs
PLeVasseur: Hmmm. What about if the Request message contained some identifying information and the UListener impl was set up so that it could use that identifying information to sleep request_A, while not doing so for request_B?

PLeVasseur (approved):
LGTM! Thanks for the updates @evshary :slightly_smiling_face:

src/utransport.rs
evshary: Then we should make sure the callback won't block while accepting the first request.
Maybe related to [the issue](https://github.com/eclipse-uprotocol/up-spec/issues/104)

stevenhartley: @evshary there will be an issue if you do not pass reqid in the zenoh message we will have issues correlating a message across transports. Is the reqid passed in the UMessage returned from the transport??

evshary: reqid is still passed in UAttributes (in UMessage). It's just not used in Zenoh mechansim.

evshary: @PLeVasseur Good diagram! I think this is correct.

PR #28, opened 2024-04-19
Fix timeout issue in RPC

evshary:
In Zenoh, if the query doesn't be destroyed after replying, the queryable will reply with timeout.
While send_response, we should also remove the query from HashMap.


PLeVasseur (approved):
@evshary -- gosh, I must have missed this. Thank you for the fix.

Issue #25, opened 2024-04-10
Add the ability to queue Zenoh callbacks


stevenhartley:
Zenoh callbacks are invoked from the same receive thread which means if a callback to the application is taking too long (or hung), the receive thread will also be hung. This PR is to address this issue (discussed during the weekly Zenoh call https://docs.google.com/document/d/1pK4Lx3T52zUD4z35ryVrZqMB6hrg3yVyKk1yaiFys3w/edit) .

evshary:
Related to [this issue](https://github.com/eclipse-uprotocol/up-spec/issues/104)

evshary:
Fix in #30

PR #30, opened 2024-04-24
Support queuing user callback and not to block Zenoh callback

evshary:
To solve #25, I added a thread pool for handling user callback. Zenoh callback will not be blocked now.


evshary:
Before I update Zenoh to 0.11.0, I'll want to merge this first.
@PLeVasseur Would you mind help me reviewing it?

src/lib.rs
```diff
@@ -100,23 +138,29 @@ impl UPClientZenoh {
...
             ..Default::default()
         };
+        // Create Zenoh session
+        let Ok(session) = zenoh::open(config).res().await else {
+            let msg = "Unable to open Zenoh session".to_string();
+            log::error!("{msg}");
+            return Err(UStatus::fail_with_code(UCode::INTERNAL, msg));
+        };
+        // Create channels for passing user callback
+        let (cb_sender, cb_receiver) = unbounded::<CallbackChannelMessage>();
```
PLeVasseur, line 153: I've read in other places that using unbounded channels could be considered an anti-pattern and that channel queues should not be allowed to grow unbounded, generally.

I don't have a ton of experience, so I've been following that advice and using bounded channels.

Could you go into a little bit of the reasoning behind making this channel unbounded?

src/lib.rs
```diff
@@ -67,29 +102,32 @@ impl UPClientZenoh {
...
+    ///     .await
+    ///     .unwrap();
     /// # });
     /// ```
     pub async fn new(
         config: Config,
         uauthority: UAuthority,
         uentity: UEntity,
     ) -> Result<UPClientZenoh, UStatus> {
+        // Validate th UUri
```
PLeVasseur, line 130:
```suggestion
        // Validate the UUri
```

src/utransport.rs
```diff
@@ -25,6 +25,32 @@ use zenoh::{
     queryable::Query,
 };
 
+macro_rules! error_callback {
```
PLeVasseur, line 28: Cool uses of `macro_rules!` :slightly_smiling_face:

src/lib.rs
```diff
@@ -14,14 +14,17 @@
 pub mod rpc;
 pub mod utransport;
 
+use async_std::task::block_on;
+use crossbeam_channel::{unbounded, Receiver, Sender};
```
PLeVasseur, line 18: I'm curious on the choice of using `crossbeam_channel` here over leaning on an async channel from `async_std`.

My guess is that since Zenoh is going to switch to using `tokio` in the 1.0 release you wanted something async framework-agnostic.

src/lib.rs
```diff
@@ -32,6 +35,36 @@ use zenoh::{
 };
 
 const UATTRIBUTE_VERSION: u8 = 1;
+const THREAD_NUM: usize = 10;
+
+type CallbackChannelMessage = (Arc<dyn UListener>, Result<UMessage, UStatus>);
+
+struct CallbackThreadPool {
```
PLeVasseur, line 42: @evshary, @sophokles73, @AnotherDaniel --

There's a [discussion](https://github.com/eclipse-uprotocol/up-cpp/pull/99#discussion_r1580105022) going on over in cpp-land during their API redesign on whether a common utility like a thread pool implementation should be offered by `up-cpp` so that all `up-client-foo-cpp` can benefit.

What do we think about offering the same for `up-rust`?

Frankly, I'm kinda not in favor of offering the same in `up-rust` because setting up a thread pool is _much_ easier in Rust than in C++. (~40 lines of code in Rust using very standard betteries-included crates, ~200 lines of code in C++)

I do think we should offer guidance within `up-rust` to implementors of `UTransport` to avoid blocking on send at the very minimum and improve the example at the bottom of `utransport.rs` to include use of a thread-pool.

Curious to hear what y'all think though.

src/lib.rs
```diff
@@ -32,6 +35,36 @@ use zenoh::{
 };
 
 const UATTRIBUTE_VERSION: u8 = 1;
+const THREAD_NUM: usize = 10;
```
PLeVasseur, line 38: Should this be configurable by an end-user or is this prone to mis-use if exposed? Certainly we don't want a user to accidentally configure a thread-pool of 100 instead of 10 :sweat_smile: 

Curious to hear your thoughts too @stevenhartley, @sophokles73, @AnotherDaniel

PLeVasseur (changes requested):
IMHO we should add some new test code which "proves" that we no longer block.

WDYT?

src/lib.rs
evshary: Hmm... bounded channel is indeed more performant, but I'm not sure whether blocking the callback when the queue is full is what we expect or not. If blocking Zenoh callback is acceptable under some conditions, maybe we can move to the bounded channel.

src/lib.rs
evshary: Good catch!

src/utransport.rs
evshary: After thinking twice, I found that maybe I can use inline now since there is no type-related issue anymore.

src/lib.rs
evshary: I second it. It's not too difficult to do the same thing in Rust. We just need to let developers know they need to avoid blocking callbacks.

src/lib.rs
evshary: The reason I used `crossbeam_channel` is that the receiving side is ThreadPool and no need to handle it as futures.

evshary:
> IMHO we should add some new test code which "proves" that we no longer block.
> 
> WDYT?

Done. I created a test for it.

src/utransport.rs
```diff
@@ -25,6 +26,41 @@ use zenoh::{
...
+#[inline]
+fn error_callback(
+    sender: &Sender<CallbackChannelMessage>,
+    listener: &Arc<dyn UListener>,
+    msg: &str,
+) {
+    log::error!("{msg}");
+    if sender
+        .send(CallbackChannelMessage {
+            listener: listener.clone(),
```
PLeVasseur, line 38: Some advice I read to make APIs more explanatory is that if we're always going to clone an Arc inside a function, we might as well ask for `Arc<T>` instead of `&Arc<T>`.

I'm sure it doesn't fit everywhere, but I'd be interested to hear your thoughts.

src/lib.rs
PLeVasseur: Bumping for visibility to @stevenhartley, @sophokles73, @AnotherDaniel

src/lib.rs
```diff
@@ -37,7 +37,10 @@ use zenoh::{
 const UATTRIBUTE_VERSION: u8 = 1;
 const THREAD_NUM: usize = 10;
 
-type CallbackChannelMessage = (Arc<dyn UListener>, Result<UMessage, UStatus>);
+struct CallbackChannelMessage {
```
PLeVasseur, line 40: I like how the resulting code looks which interacts with `CallbackChannelMessage` now that it's a struct :+1:

src/lib.rs
```diff
@@ -54,8 +54,9 @@ impl CallbackThreadPool {
         let mut thread_pool = Vec::with_capacity(thread_num);
         for _ in 0..thread_num {
             let receiver = cb_receiver.clone();
-            thread_pool.push(thread::spawn(move || loop {
-                if let Ok(msg) = receiver.recv() {
+            thread_pool.push(thread::spawn(move || {
+                // if cb_sender is released, break the while-loop and stop the thread
+                while let Ok(msg) = receiver.recv() {
```
PLeVasseur, line 59: Practically speaking, this will only occur if `UPClientZenoh` has been dropped, right? Seems reasonable to me.

src/lib.rs
PLeVasseur: :thinking: 

From what I'm reading on best practices, generally speaking, it's better to use bounded channels and explicitly consider what should occur when the channel's full:
* block
* lose messsages

IMHO losing messages is not appropriate, so as you said we should probably block.

Reading suggestions that the blocking behavior creating back-pressure is a _good thing_ since it's explicit.

In a similar question to [this one](https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/30#discussion_r1581046393), I'm wondering if the size of the channel should be explicitly set by a user or something that's set to "a large enough number".

Should we:
1. set to some "large" number relative to typical use-case and call it good for now or
2. expose the parameter?

PLeVasseur:
> Done. I created a test for it.

The test looks good :)

src/lib.rs
evshary: You're right. While `UPClientZenoh` is dropped, the `cb_sender` will be released and also break the loop here to stop threads.

src/utransport.rs
evshary: This is what clippy asks me to do :laughing: 

```
error: this argument is passed by value, but not consumed in the function body
  --> src/utransport.rs:50:15
   |
50 |     listener: Arc<dyn UListener>,
   |               ^^^^^^^^^^^^^^^^^^ help: consider taking a reference instead: `&Arc<dyn UListener>`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value
```

I guess the clippy doesn't think it necessary to bump up /down the reference count here.
Here is the explanation from others:
https://stackoverflow.com/questions/55576425/why-does-clippy-suggests-passing-an-arc-as-a-reference

PLeVasseur:
I think essentially we're in good shape aside from the two design questions I brought up.

Anyone that could weigh in would be appreciated.

evshary:
> 1. leave channel capacity as unbounded or
> 2. set to some "large" number relative to typical use-case and call it good for now or
> 3. expose the parameter?

@PLeVasseur After thinking twice, I think we can go with the second choice. A large size queue shouldn't be full in most cases, and it can keep the performance. If it's unfortunately full, blocking in this rare case should be acceptable.

As for exposing the parameters of queue size and thread number, I think we can see how users use it before making the decision. Exposing more parameters means more complicated for users to use.

PLeVasseur (approved):
> As for exposing the parameters of queue size and thread number, I think we can see how users use it before making the decision. Exposing more parameters means more complicated for users to use.

Yeah, I think this is fair. Let's expose them on an as-needed basis.

Since that was the last remaining item, approved! :slightly_smiling_face:

PR #33, opened 2024-05-07
Bump Zenoh to 0.11.0

evshary:
The PR is to update Zenoh to 0.11.0, and also use `tokio` instead of `async_std`.


Cargo.toml
```diff
@@ -23,7 +23,8 @@ prost-types = "0.12"
 protobuf = { version = "3.3" }
 rand = "0.8.5"
 up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", rev = "1bb08ba7a3666e58c316489fbcf3da3e29dee611" }
-zenoh = { version = "0.10.1-rc", features = ["unstable"]}
+zenoh = { git = "https://github.com/eclipse-zenoh/zenoh", rev = "b8dd01d6ce05f851f33497e8d899df308d315afb", features = ["unstable"] }
```
PLeVasseur, line 26: No need for a tonne of detail.

A part of me is curious as to why the released version isn't chosen though 🤔

src/lib.rs
```diff
@@ -56,9 +55,11 @@ impl CallbackThreadPool {
         for _ in 0..thread_num {
             let receiver = cb_receiver.clone();
             thread_pool.push(thread::spawn(move || {
+                // Create a dedicated runtime for the thread
+                let rt = tokio::runtime::Runtime::new().expect("Unable to create runtime");
```
PLeVasseur, line 59: I suppose this was the case implicitly earlier using async-std as well, seems like we're starting a new tokio runtime per thread.

I haven't used tokio much -- is this heavy? Standard practice when wanting a dedicated async thread?

src/utransport.rs
```diff
@@ -27,37 +27,53 @@ use zenoh::{
...
-            result: Err(UStatus::fail_with_code(UCode::INTERNAL, msg)),
-        })
-        .is_err()
-    {
-        log::error!("Unable to call the user callback");
+fn invoke_block_callback(listener: &Arc<dyn UListener>, resp_msg: Result<UMessage, &str>) {
+    match resp_msg {
+        Ok(umsg) => {
+            task::block_in_place(|| {
+                Handle::current().block_on(listener.on_receive(umsg));
```
PLeVasseur, line 34: tokio is so much more explicit! Interesting.

So we need to grab a handle to the tokio Runtime running this function so that we can call block_on.

src/utransport.rs
```diff
@@ -27,37 +27,53 @@ use zenoh::{
...
+                .send(CallbackChannelMessage {
+                    listener: listener.clone(),
+                    result: Ok(umsg),
+                })
+                .is_err()
+            {
+                log::error!("Unable to call the user callback");
+            }
+        }
+        Err(msg) => {
```
PLeVasseur, line 65:
```suggestion
        Err(err) => {
```
?

src/lib.rs
PLeVasseur, line 59: I think this is a reasonable place to panic. 👍

The developer using UPClient can't really do anything if unable to create a tokio Runtime

src/utransport.rs
```diff
@@ -27,37 +27,53 @@ use zenoh::{
...
+            log::error!("{msg}");
+            if sender
+                .send(CallbackChannelMessage {
+                    listener: listener.clone(),
+                    result: Err(UStatus::fail_with_code(UCode::INTERNAL, msg)),
+                })
+                .is_err()
+            {
+                log::error!("Unable to call the user callback");
```
PLeVasseur, line 74: What do you think about adding a little more detail as to why?

Probably will help tracking down bugs later in developer code.

src/utransport.rs
```diff
@@ -27,37 +27,53 @@ use zenoh::{
...
+            }
+        }
+        Err(msg) => {
+            log::error!("{msg}");
+            if sender
+                .send(CallbackChannelMessage {
+                    listener: listener.clone(),
+                    result: Err(UStatus::fail_with_code(UCode::INTERNAL, msg)),
```
PLeVasseur, line 70: What about [`RESOURCE_EXHAUSTED`](https://github.com/eclipse-uprotocol/up-spec/blob/19bec77b3c3222def74604dbe5f5d0af2afb9ae1/up-core-api/uprotocol/ustatus.proto#L84) instead? Seems more fitting to me

src/utransport.rs
```diff
@@ -152,37 +168,30 @@ impl UPClientZenoh {
                 Ok(sample) => {
                     // Get the encoding of UPayload
                     let Some(encoding) = UPClientZenoh::to_upayload_format(&sample.encoding) else {
-                        let msg = "Unable to get the encoding".to_string();
-                        log::error!("{msg}");
-                        block_on(
-                            resp_callback.on_error(UStatus::fail_with_code(UCode::INTERNAL, msg)),
-                        );
+                        invoke_block_callback(&resp_callback, Err("Unable to get the encoding"));
```
PLeVasseur, line 171: Should this be blocking? 🤔

Could it be non-blocking? What if I want to send out N requests one after another in the uStreamer.

PLeVasseur (changes requested):
Overall looks good

Left some questions to help my understanding

Cargo.toml
evshary: In fact, there are some issues in 0.11.0-rc.2, so I want to include the bugfix to avoid CI failure.
Now 0.11.0-rc.3 is released. I'll switch to tha.

src/lib.rs
evshary: IMHO, if we use the thread pool to handle the callback, I don't think we need async anymore. However, the callback is async func, so it forces us to create a runtime to run the async func.
In fact, I'm thinking about whether using thread pool here is a good idea or not. Maybe we should have a dedicated runtime here (with multithreaded config) instead of running async func in every single thread.

src/utransport.rs
evshary: Correct. In async_std, we don't need to specify runtime and can use block_on directly. However, in tokio, we can't call nested block_on. It forces us to use block_in_place and specify the handle explicitly.

src/utransport.rs
evshary: In fact, this does nothing to your requests since this is inside the zenoh_callback, which means you can call several requests without blocking.

src/lib.rs
evshary: Yes, I don't like panic, but it seems like I can do nothing here...

PLeVasseur: > Maybe we should have a dedicated runtime here (with multithreaded config) instead of running async func in every single thread.

:thinking: Honestly, I dunno which is more performant or the trade-offs.

Maybe list up the trade-offs and then do some relatively quick testing?

src/utransport.rs
PLeVasseur: Gotcha. That makes sense.

PLeVasseur (approved):
LGTM

If you'd like to do the testing we discussed of threadpool of tokio runtimes vs tokio multi-threaded runtime and want further review, let me know.

evshary:
> If you'd like to do the testing we discussed of threadpool of tokio runtimes vs tokio multi-threaded runtime and want further review, let me know.

I'm replacing threadpool with mutlit-threaded runtime, but still blocked by some issues.
I will bump up-rust first, so it would be better to include Zenoh 0.11.0 in code. I'll merge the code first and do the test in the following PR.

PR #34, opened 2024-05-10
Update the license headers

evshary:
@stevenhartley I updated the license header to be consistent with up-rust.


PLeVasseur (approved):
LGTM

PR #38, opened 2024-05-20
Use Runtime with worker threads to replace thread pool.

evshary:
Originally, we created thread pools and generated runtime for each thread to spawn async callback. However, this is not efficient and doesn't utilize the power of async. Now I create a separate runtime to spawn the callbacks. We can specify the thread number of the runtime and the runtime can asynchronously handle these callbacks.

@PLeVasseur This is the follow-up PR of #33


PLeVasseur (approved):
LGTM

Issue #39, opened 2024-05-31
Bump the up-rust to 1.5.8


evshary:
Still WIP. Just create this issue for others to track.

PR #40, opened 2024-06-03
Initial support 1.5.8

evshary:
The PR is to support up-spec 1.5.8 
Close https://github.com/eclipse-uprotocol/up-client-zenoh-rust/issues/39

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,16 +13,14 @@
 pub mod rpc;
 pub mod utransport;
 
-use protobuf::{Enum, Message};
+use bitmask_enum::bitmask;
+use protobuf::Message;
 use std::{
     collections::HashMap,
     sync::{Arc, Mutex},
 };
 use tokio::runtime::Runtime;
-use up_rust::{
-    ComparableListener, UAttributes, UAuthority, UCode, UEntity, UListener, UPayloadFormat,
-    UPriority, UResourceBuilder, UStatus, UUri,
-};
+use up_rust::{ComparableListener, UAttributes, UCode, UListener, UPriority, UStatus, UUri};
 use zenoh::{
     config::Config,
     prelude::r#async::*,
@@ -30,6 +28,15 @@ use zenoh::{
     sample::{Attachment, AttachmentBuilder},
     subscriber::Subscriber,
 };
+
+// CY_TODO: Whether to expose from up_rust or not
+const _WILDCARD_AUTHORITY: &str = "*";
+const WILDCARD_ENTITY_ID: u32 = 0x0000_FFFF;
+const WILDCARD_ENTITY_VERSION: u32 = 0x0000_00FF;
+const WILDCARD_RESOURCE_ID: u32 = 0x0000_FFFF;
+
+const _RESOURCE_ID_RESPONSE: u32 = 0;
+const _RESOURCE_ID_MIN_EVENT: u32 = 0x8000;
 
 const UATTRIBUTE_VERSION: u8 = 1;
 const THREAD_NUM: usize = 10;
@@ -43,10 +50,18 @@ lazy_static::lazy_static! {
                .expect("Unable to create callback runtime");
 }
 
-type SubscriberMap = Arc<Mutex<HashMap<(UUri, ComparableListener), Subscriber<'static, ()>>>>;
-type QueryableMap = Arc<Mutex<HashMap<(UUri, ComparableListener), Queryable<'static, ()>>>>;
+#[bitmask(u8)]
+enum MessageFlag {
+    Publish,
+    Notification,
+    Request,
+    Response,
+}
+
+type SubscriberMap = Arc<Mutex<HashMap<(String, ComparableListener), Subscriber<'static, ()>>>>;
+type QueryableMap = Arc<Mutex<HashMap<(String, ComparableListener), Queryable<'static, ()>>>>;
 type QueryMap = Arc<Mutex<HashMap<String, Query>>>;
-type RpcCallbackMap = Arc<Mutex<HashMap<UUri, Arc<dyn UListener>>>>;
+type RpcCallbackMap = Arc<Mutex<HashMap<String, Arc<dyn UListener>>>>;
 pub struct UPClientZenoh {
     session: Arc<Session>,
     // Able to unregister Subscriber
@@ -57,8 +72,8 @@ pub struct UPClientZenoh {
     query_map: QueryMap,
     // Save the callback for RPC response
     rpc_callback_map: RpcCallbackMap,
-    // Source UUri in RPC
-    source_uuri: UUri,
+    // My authority
+    authority_name: String,
 }
 
 impl UPClientZenoh {
@@ -66,9 +81,8 @@ impl UPClientZenoh {
     ///
     /// # Arguments
     ///
-    /// * `config` - Zenoh configuration. You can refer to [here](https://github.com/eclipse-zenoh/zenoh/blob/0.10.1-rc/DEFAULT_CONFIG.json5) for more configuration details.
-    /// * `uauthority` - The `UAuthority` which is put into source address while generating messages.
-    /// * `uentity` - The `UEntity` which is put into source address while generating messages.
+    /// * `config` - Zenoh configuration. You can refer to [here](https://github.com/eclipse-zenoh/zenoh/blob/0.11.0-rc.3/DEFAULT_CONFIG.json5) for more configuration details.
+    /// * `authority_name` - The authority name. We need it to generate Zenoh key since authority might be omitted in `UUri`.
     ///
     /// # Errors
     /// Will return `Err` if unable to create `UPClientZenoh`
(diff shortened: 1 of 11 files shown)


evshary:
@PLeVasseur Would you mind taking a look?

PLeVasseur:
Hi @evshary -- I'm on vacation starting tomorrow till next Monday. I'll see if I can sneak a review of this in. :)

evshary:
> Hi @evshary -- I'm on vacation starting tomorrow till next Monday. I'll see if I can sneak a review of this in. :)

Hi @PLeVasseur It's fine to be reviewed next week. Enjoy your vacation!

src/lib.rs
```diff
@@ -31,6 +29,15 @@ use zenoh::{
     subscriber::Subscriber,
 };
 
+// CY_TODO: Whether to expose from up_rust or not
```
PLeVasseur, line 32: Very good question. IMHO it makes sense for these to live in `up-rust`.

src/lib.rs
```diff
@@ -277,97 +208,111 @@ impl UPClientZenoh {
         };
         Ok(uattributes)
     }
+
+    /*        The table for mapping resource ID to message type
```
PLeVasseur, line 212: If this table doesn't exist in up-spec it'd be great to have you contribute to adding it there!

src/utransport.rs
```diff
@@ -534,121 +446,117 @@ impl UTransport for UPClientZenoh {
...
     ) -> Result<(), UStatus> {
-        if topic.authority.is_some() && topic.entity.is_none() && topic.resource.is_none() {
-            // This is special UUri which means we need to register for all of Publish, Notification, Request, and Response
-            // RPC response
-            self.register_response_listener(&topic, listener.clone());
-            // RPC request
-            self.register_request_listener(&topic, listener.clone())
+        let flag = UPClientZenoh::get_listener_message_type(source_filter, sink_filter)?;
+        // Publish & Notification
+        if flag.contains(MessageFlag::Publish) || flag.contains(MessageFlag::Notification) {
```
PLeVasseur, line 467: So I think based on how I am reading this, the intention looks to be that the source and sink filters may result in more than one bit being set in the bitmask enum. That seems to be the intent. And that makes sense.

I'm having a hard time seeing how the function get_listener_message_type() can set multiple bits. It looks like a single value is set based on the if conditionals and returned. Hmm. Perhaps I need to read more 🤔

tests/register.rs
```diff
@@ -25,32 +25,37 @@ impl UListener for FooListener {
     async fn on_error(&self, _err: UStatus) {}
 }
 
-#[test_case(test_lib::create_utransport_uuri(Some(0), 0, 0); "Publish / Notification register_listener")]
-#[test_case(test_lib::create_rpcserver_uuri(Some(0), 0); "RPC register_listener")]
-#[test_case(test_lib::create_special_uuri(0); "Special UUri register_listener")]
+#[test_case(&test_lib::new_uuri("src", 1, 1, 0x8000), None; "Listen to Publish")]
```
PLeVasseur, line 28: Is there a way to listen to all point to point messages? I see ability to listen to all messages _of_ a given type. What about being able to express "all messages with a sink whose source authority_name is foo"?

That's one of the streamer use cases for connecting two authorities together over two transports.

PLeVasseur: Just checked the uP-L1 Zenoh spec. Looks like this case may have been missed. Can we update the spec to clarify this is also possible and add this capability to the transport? (If I am reading it correctly it looks like registering to get all point to point messages as it was in 1.5.7 is not possible in the current implementation. Very possible I'm wrong and if so please point it out 😅)

src/lib.rs
```diff
@@ -277,97 +208,111 @@ impl UPClientZenoh {
...
+     *  |     FFFF    |   FFFF   |         |      V       |    V    |    V     |
+     *
+     *  Some organization:
+     *  - Publish: {[8000-FFFF), None}
+     *  - Notification: {[8000-FFFF), 0}, {[8000-FFFF), FFFF]}, {FFFF, 0}, {FFFF, FFFF}
+     *  - Request: {0, (0-8000)}, {0, FFFF}, {FFFF, (0-8000)}, {FFFF, FFFF}
+     *  - Response: {(0-8000), 0}, {(0-8000), FFFF}, (FFFF, 0), {FFFF, FFFF}
+     */
+    #[allow(clippy::nonminimal_bool)] // Don't simplify the boolean expression for better understanding
+    fn get_listener_message_type(
```
PLeVasseur, line 234: IMHO something very similar could be useful in up-rust with modifications to support the uStreamer case of all point to point messages from a certain authority.

src/lib.rs
evshary: So is the table clear to you? The reason why I hesitate to add the table into spec is that I'm afraid it makes readers difficult to understand.

PLeVasseur: Yeah, I get it. I spent a few minutes just mentally going over the table to see if it aligned with my understanding.

I think, even if it takes some time for someone to process what's in the table, it makes a lot more sense to have it written up in a place like up-spec that can be referenced for any implementation :slightly_smiling_face:

src/utransport.rs
evshary: You're right... I made the silly mistakes.
I updated how `get_listener_message_type` works and added more tests case to avoid the mistakes in the future.

tests/register.rs
evshary: Is this what you expect? I added the uStreamer test cases in the tests now.
https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/pull/40/commits/ea3a993497967418c31060ddeed91ea9fb632f5a#diff-c19a9c920bb76a5ab06a7d28f16df025d34c3a66697465b9e92c9df37ecdc0a3R36

PLeVasseur: I think this is pretty close, left [a comment](https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/commit/ea3a993497967418c31060ddeed91ea9fb632f5a#r143069220) on that test.

PLeVasseur (changes requested):
I think you may need to modify or make a new so that we can test the uStreamer use-case of all of Notification, Request, and Response.

PLeVasseur:
I think it'd be good to update the name to `up-transport-zenoh`, too.

PLeVasseur:
As I start to work with the Zenoh transport, I think it would be good if from this crate we re-exported some key Zenoh items such that an end-user of up-transport-zenoh-rust doesn't _also_ need to pull in Zenoh as a dependency and can use the stuff required from there.

I'm thinking specifically at the moment about Zenoh's Config, as it would appear in order to configure the Zenoh transport I do need to pass in a `Config` into `new()`.

src/lib.rs
```diff
@@ -13,16 +13,14 @@
...
     sync::{Arc, Mutex},
 };
 use tokio::runtime::Runtime;
-use up_rust::{
-    ComparableListener, UAttributes, UAuthority, UCode, UEntity, UListener, UPayloadFormat,
-    UPriority, UResourceBuilder, UStatus, UUri,
-};
+use up_rust::{ComparableListener, UAttributes, UCode, UListener, UPriority, UStatus, UUri};
 use zenoh::{
     config::Config,
```
PLeVasseur, line 25: Suggest re-exporting certain key things from the Zenoh crate, maybe start with Config?

evshary: Good idea

evshary:
> I think you may need to modify or make a new test so that we can test the uStreamer use-case of all of Notification, Request, and Response.

I found something wrong with Response use case. In the previous implementation, calling `register_listener` for Response message doesn't work with a wildcard. It always needs to have the fully-matching Zenoh key. That means the Zenoh key from `register_listener` and `send(request)` should be the same.

I fixed the issue with the help of the intersection in Zenoh key. Now `register_listener` and `send(request)` don't need to be the same. They can be intersected.

@PLeVasseur Let me know if the test case is not what you imagine.

src/lib.rs
```diff
@@ -78,47 +93,13 @@ impl UPClientZenoh {
...
-            let msg = format!("UEntity is invalid: {e:?}");
-            log::error!("{msg}");
-            UStatus::fail_with_code(UCode::INVALID_ARGUMENT, msg)
-        })?;
-        let source_uuri = UUri {
-            authority: Some(uauthority).into(),
-            entity: Some(uentity).into(),
-            ..Default::default()
-        };
+    pub async fn new(config: Config, authority_name: String) -> Result<UPClientZenoh, UStatus> {
```
PLeVasseur, line 102: Hey @evshary -- for the uStreamer use-case we'll need a function to build a UPClientZenoh object from a Zenoh Runtime in order to fit within the topology with up-linux-streamer as a Zenoh plugin.

I'm okay if you want to merge this one and address that in a follow up. :+1:

src/lib.rs
```diff
@@ -21,8 +21,9 @@ use std::{
 };
 use tokio::runtime::Runtime;
 use up_rust::{ComparableListener, UAttributes, UCode, UListener, UPriority, UStatus, UUri};
+// Re-export Zenoh config
+pub use zenoh::config::Config;
```
PLeVasseur, line 25: Is this enough to configure a Config completely? Or should we re-export all of zenoh::config::*?

evshary: It seems like there will be some conflicts if expose all config. I'll deal with it later.

src/lib.rs
evshary: OK. Don't make the PR to lengthy. I'll create an issue to track.

PR #47, opened 2024-06-17
Bump up-rust for UCode .proto fix

PLeVasseur:
Bumps to a version of up-rust with fix for UCode now being its own .proto


evshary (approved):
LGTM

Issue #42, opened 2024-06-14
Support initializing with Zenoh Runtime


evshary:
As mentioned [here](https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/pull/40#discussion_r1638275529), we should support using Zenoh Runtime to do the construction.

evshary:
Close #49

PR #49, opened 2024-06-21
Support init with Runtime

evshary:
@PLeVasseur As mentioned in #42, I created a constructor to accept Runtime. It seems like the only way to create Runtime is to use plugins, so I can't write tests for it. Let me know if you have any issues while using it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,6 +26,7 @@ pub use zenoh::config::Config;
 use zenoh::{
     prelude::r#async::*,
     queryable::{Query, Queryable},
+    runtime::Runtime as ZRuntime,
     sample::{Attachment, AttachmentBuilder},
     subscriber::Subscriber,
 };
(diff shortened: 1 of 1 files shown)


evshary:
@PLeVasseur By the way, I'm not sure why I can't request you as a reviewer on GitHub...

PLeVasseur:
I think you can only request committers, and while it got approved through Eclipse I am probably not yet a committer on the project on GitHub.

Just my guess tho 🤔

PLeVasseur (approved):
Hey @evshary -- this works! Got it working when run inside of a Zenoh Plugin.

Can we merge it so I can rely on main for the demo?

There is a part of me that wonders... is there no way to create a Zenoh Runtime outside of being a Plugin? I feel like it should be possible in order to unit test all the Plugin stuff in mainline Zenoh? :thinking:

evshary:
@PLeVasseur Sure. I took a quick look at the unit test. It seems like unit test focuses on other functions (not the full initialization). There is also an example plugin, and I guess it kind of covers the test of initialization.

PR #50, opened 2024-06-26
Added handling of the case where we're within a Tokio Runtime or not

PLeVasseur:
# What

Added handling of the case where we're within a Tokio Runtime or not
* if we're within a Tokio Runtime use the current handle
* if not, use a dedicated Tokio Runtime we have (e.g. we're within an async-std executor)

# Why

To handle the current case / issue of trying to send a Request UMessage within a Zenoh Plugin on zenoh 0.11.0-rc3, which I _think_ still uses async-std. Therefore we need to ensure we use a dedicated Tokio Runtime in that case within up-transport-zenoh-rust.

With this change, we resolve [this issue](https://github.com/eclipse-uprotocol/up-streamer-rust/pull/17#:~:text=Issue%20with%20up%2Dtransport%2Dzenoh%2Drust%20assuming%20it%27s%20inside%20of%20a%20Tokio%20runtime%3F) pointed out on the up-linux-streamer-plugin PR:
> Issue with up-transport-zenoh-rust assuming it's inside of a Tokio runtime?

### DIFF
--- a/src/utransport.rs
+++ b/src/utransport.rs
@@ -13,10 +13,14 @@
 use crate::{MessageFlag, UPClientZenoh, CB_RUNTIME};
 use async_trait::async_trait;
 use lazy_static::lazy_static;
-use std::sync::Mutex;
-use std::{sync::Arc, time::Duration};
-use tokio::runtime::Runtime;
-use tokio::{runtime::Handle, task};
+use std::{
+    sync::{Arc, Mutex},
+    time::Duration,
+};
+use tokio::{
+    runtime::{Handle, Runtime},
+    task,
+};
 use up_rust::{
     ComparableListener, UAttributes, UAttributesValidators, UCode, UListener, UMessage,
     UMessageType, UStatus, UTransport, UUri,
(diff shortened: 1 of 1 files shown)


src/utransport.rs
```diff
@@ -12,7 +12,10 @@
  ********************************************************************************/
 use crate::{MessageFlag, UPClientZenoh, CB_RUNTIME};
 use async_trait::async_trait;
+use lazy_static::lazy_static;
+use std::sync::Mutex;
 use std::{sync::Arc, time::Duration};
+use tokio::runtime::Runtime;
```
evshary, line 18: I prefer to merge the crate path together.

evshary (commented):
Interesting. I didn't find that the plugins in zenoh are still using async_std. Also, my assumption of using tokio might be wrong. I'm ok with the current change, but maybe there is a better way to be compatible with both tokio and aync_std. Let's figure it out later.

evshary:
@PLeVasseur I know you might need the PR urgently for the uStreamer demo. I take the liberty to update the PR and merge it. I hope it doesn't offend you.

PR #51, opened 2024-07-08
Use tracing to replace log

evshary:
Now up-rust and Zenoh are both using tracing instead of log. Migrate the log system to ensure consistency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,14 +40,14 @@ async-trait = "0.1"
 bitmask-enum = "2.2.4"
 chrono = "0.4.31"
 crossbeam-channel = "0.5.12"
-env_logger = "0.10.0"
 lazy_static = "1.4.0"
-log = "0.4.17"
 prost = "0.12"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -40,14 +40,14 @@ async-trait = "0.1"
...
 crossbeam-channel = "0.5.12"
-env_logger = "0.10.0"
 lazy_static = "1.4.0"
-log = "0.4.17"
 prost = "0.12"
 prost-types = "0.12"
 protobuf = { version = "3.3" }
 rand = "0.8.5"
 tokio = { version = "1.35.1", default-features = false }
+tracing = "0.1.40"
```
PLeVasseur, line 49: I realized as well that this is the superior crate to use for multi-threaded and very-async libraries

Want to do the same for up-transport-vsomeip-rust and up-streamer-rust in a later PR :slightly_smiling_face:

PLeVasseur (approved):
Yup! A great switch

PR #52, opened 2024-07-08
Re-export all the Zenoh config.

evshary:
Deal with https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/issues/43

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,7 +22,7 @@ use std::{
 use tokio::runtime::Runtime;
 use up_rust::{ComparableListener, UAttributes, UCode, UListener, UPriority, UStatus, UUri};
 // Re-export Zenoh config
-pub use zenoh::config::Config;
+pub use zenoh::config as ZenohConfig;
 use zenoh::{
     prelude::r#async::*,
     queryable::{Query, Queryable},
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -22,7 +22,7 @@ use std::{
 use tokio::runtime::Runtime;
 use up_rust::{ComparableListener, UAttributes, UCode, UListener, UPriority, UStatus, UUri};
 // Re-export Zenoh config
-pub use zenoh::config::Config;
+pub use zenoh::config as ZenohConfig;
```
PLeVasseur, line 25: Since you're aliasing a module here, would this be better as `zenoh_config`?

src/lib.rs
```diff
@@ -94,13 +94,16 @@ impl UPClientZenoh {
     /// ```
     /// #[tokio::main]
     /// # async fn main() {
-    /// use up_transport_zenoh::{Config, UPClientZenoh};
-    /// let upclient = UPClientZenoh::new(Config::default(), String::from("MyAuthName"))
+    /// use up_transport_zenoh::{UPClientZenoh, ZenohConfig};
+    /// let upclient = UPClientZenoh::new(ZenohConfig::Config::default(), String::from("MyAuthName"))
```
PLeVasseur, line 98: For example, if this were `zenoh_config::Config::default()` here this would read more naturally to me in Rust.

PLeVasseur (changes requested):
Suggested an update from `ZenohConfig` to `zenoh_config` since this is a module alias.

src/lib.rs
evshary: Haha, I struggled to choose between `ZenohConfig` and `zenoh_config` while using the alias name. I chose `zenoh_config` first and then renamed it to `ZenohConfig`. I'm not good at naming, so I'll follow your preference :laughing:

PLeVasseur (approved):
LGTM

PR #56, opened 2024-07-16
Update to the latest up-rust (Support L2 API)

evshary:
The PR is to use the latest up-rust, including the following update:

1. remove `on_error` in UListener
2. Support ZenohRpcClient (New way for L2 API)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ rand = "0.8.5"
 tokio = { version = "1.35.1", default-features = false }
 tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
-up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", rev = "3a50104421a801d52e1d9c68979db54c013ce43d" }
+up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", rev = "a74c829d3e54ce09c7b85ee721be7e8ac703aac9" }
 zenoh = { version = "0.11.0-rc.3", features = ["unstable"]}
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ tokio = { version = "1.35.1", default-features = false }
 tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
 up-rust = "0.1.3"
-zenoh = { version = "0.11.0-rc.3", features = ["unstable"]}
+zenoh = { version = "0.11.0", features = ["unstable"]}
 
 [dev-dependencies]
 test-case = { version = "3.3" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ rand = "0.8.5"
 tokio = { version = "1.35.1", default-features = false }
 tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
-up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", rev = "cf5af235f4fa267bdacf7e313347708c64a11ac1" }
+up-rust = "0.1.3"
 zenoh = { version = "0.11.0-rc.3", features = ["unstable"]}
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ rand = "0.8.5"
 tokio = { version = "1.35.1", default-features = false }
 tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
-up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", rev = "a74c829d3e54ce09c7b85ee721be7e8ac703aac9" }
+up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", rev = "cf5af235f4fa267bdacf7e313347708c64a11ac1" }
 zenoh = { version = "0.11.0-rc.3", features = ["unstable"]}
 
 [dev-dependencies]
--- a/examples/rpc_client.rs
+++ b/examples/rpc_client.rs
@@ -13,11 +13,8 @@
 mod common;
 
 use async_trait::async_trait;
-use std::{
-    str::FromStr,
-    sync::{Arc, Mutex},
-};
-use tokio::time::{sleep, Duration};
+use std::{str::FromStr, sync::Arc};
+use tokio::sync::Notify;
 use up_rust::{
     LocalUriProvider, UListener, UMessage, UMessageBuilder, UPayloadFormat, UTransport, UUri,
 };
@@ -25,13 +22,11 @@ use up_transport_zenoh::UPTransportZenoh;
 
 // ResponseListener
 struct ResponseListener {
-    response: Mutex<Option<String>>,
+    notify: Arc<Notify>,
 }
 impl ResponseListener {
-    fn new() -> Self {
-        ResponseListener {
-            response: Mutex::new(None),
-        }
+    fn new(notify: Arc<Notify>) -> Self {
+        Self { notify }
     }
 }
 #[async_trait]
@@ -40,9 +35,8 @@ impl UListener for ResponseListener {
         let payload = msg.payload.unwrap();
         let value = payload.into_iter().map(|c| c as char).collect::<String>();
         let uri = msg.attributes.unwrap().source.unwrap().to_string();
-        let mut resp = self.response.lock().unwrap();
-        *resp = Some(value.clone());
         println!("Receiving response {value} from {uri}");
+        self.notify.notify_one();
     }
 }
 
@@ -61,7 +55,8 @@ async fn main() {
     let sink_uuri = UUri::from_str("//rpc_server/1/1/1").unwrap();
 
     // register response callback
-    let resp_listener = Arc::new(ResponseListener::new());
+    let notify = Arc::new(Notify::new());
+    let resp_listener = Arc::new(ResponseListener::new(notify.clone()));
     rpc_client
         .register_listener(&sink_uuri, Some(&src_uuri), resp_listener.clone())
         .await
@@ -75,7 +70,5 @@ async fn main() {
     println!("Sending request from {src_uuri} to {sink_uuri}");
     rpc_client.send(umsg).await.unwrap();
 
-    while resp_listener.response.lock().unwrap().is_none() {
-        sleep(Duration::from_millis(100)).await;
-    }
+    notify.notified().await;
 }
--- a/examples/rpc_client.rs
+++ b/examples/rpc_client.rs
@@ -70,7 +70,7 @@ async fn main() {
     let umsg = UMessageBuilder::request(sink_uuri.clone(), src_uuri.clone(), 1000)
         .build_with_payload(data, UPayloadFormat::UPAYLOAD_FORMAT_TEXT)
         .unwrap();
-    println!("Send request from {src_uuri} to {sink_uuri}");
+    println!("Sending request from {src_uuri} to {sink_uuri}");
     rpc_client.send(umsg).await.unwrap();
 
     while resp_listener.response.lock().unwrap().is_none() {
--- a/examples/rpc_client.rs
+++ b/examples/rpc_client.rs
@@ -13,7 +13,7 @@
 mod common;
 
 use std::str::FromStr;
-use up_rust::{RpcClient, UMessageBuilder, UPayloadFormat, UUri};
+use up_rust::{UMessageBuilder, UPayloadFormat, UUri};
 use up_transport_zenoh::UPClientZenoh;
 
 #[tokio::main]
@@ -22,7 +22,7 @@ async fn main() {
     UPClientZenoh::try_init_log_from_env();
 
     println!("uProtocol RPC client example");
-    let rpc_client = UPClientZenoh::new(common::get_zenoh_config(), String::from("rpc_client"))
+    let _rpc_client = UPClientZenoh::new(common::get_zenoh_config(), String::from("rpc_client"))
         .await
         .unwrap();
 
@@ -32,16 +32,16 @@ async fn main() {
 
     // create uPayload
     let data = String::from("GetCurrentTime");
-    let umsg = UMessageBuilder::request(sink_uuri.clone(), src_uuri.clone(), 1000)
+    let _umsg = UMessageBuilder::request(sink_uuri.clone(), src_uuri.clone(), 1000)
         .build_with_payload(data, UPayloadFormat::UPAYLOAD_FORMAT_TEXT)
         .unwrap();
 
-    // invoke RPC method
-    println!("Send request from {src_uuri} to {sink_uuri}");
-    let result = rpc_client.invoke_method(sink_uuri, umsg).await;
+    //// invoke RPC method
+    //println!("Send request from {src_uuri} to {sink_uuri}");
+    //let result = rpc_client.invoke_method(sink_uuri, umsg).await;
 
-    // process the result
-    let payload = result.unwrap().payload.unwrap();
-    let value = payload.into_iter().map(|c| c as char).collect::<String>();
-    println!("Receive {value}");
+    //// process the result
+    //let payload = result.unwrap().payload.unwrap();
+    //let value = payload.into_iter().map(|c| c as char).collect::<String>();
+    //println!("Receive {value}");
 }
--- a/examples/rpc_client.rs
+++ b/examples/rpc_client.rs
@@ -12,9 +12,37 @@
  ********************************************************************************/
 mod common;
 
-use std::str::FromStr;
-use up_rust::{UMessageBuilder, UPayloadFormat, UUri};
+use async_trait::async_trait;
+use std::{
+    str::FromStr,
+    sync::{Arc, Mutex},
+};
+use tokio::time::{sleep, Duration};
+use up_rust::{UListener, UMessage, UMessageBuilder, UPayloadFormat, UTransport, UUri};
 use up_transport_zenoh::UPClientZenoh;
+
+// ResponseListener
+struct ResponseListener {
+    response: Mutex<Option<String>>,
+}
+impl ResponseListener {
+    fn new() -> Self {
+        ResponseListener {
+            response: Mutex::new(None),
+        }
+    }
+}
+#[async_trait]
+impl UListener for ResponseListener {
+    async fn on_receive(&self, msg: UMessage) {
+        let payload = msg.payload.unwrap();
+        let value = payload.into_iter().map(|c| c as char).collect::<String>();
+        let uri = msg.attributes.unwrap().source.unwrap().to_string();
+        let mut resp = self.response.lock().unwrap();
+        *resp = Some(value.clone());
+        println!("Receiving response {value} from {uri}");
+    }
+}
 
 #[tokio::main]
 async fn main() {
@@ -22,7 +50,7 @@ async fn main() {
     UPClientZenoh::try_init_log_from_env();
 
     println!("uProtocol RPC client example");
-    let _rpc_client = UPClientZenoh::new(common::get_zenoh_config(), String::from("rpc_client"))
+    let rpc_client = UPClientZenoh::new(common::get_zenoh_config(), String::from("rpc_client"))
         .await
         .unwrap();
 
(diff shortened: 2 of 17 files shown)


evshary:
Since up-rust is already released, I also use up-rust 0.1.0 on crates.io here.

PLeVasseur:
I'll start to look at this!

PLeVasseur:
I saw v0.1.1 came out. I think I'll update the up-transport-vsomeip-rust to that version. Can you do the same here? I'll then go do the same for the up-streamer-rust so that we're aligned within the uStreamer.

evshary:
> I'll start to look at this!

Thank you! I always rely on your review on up-transport-zenoh-rust :smile: 

> I saw v0.1.1 came out. I think I'll update the up-transport-vsomeip-rust to that version. Can you do the same here? I'll then go do the same for the up-streamer-rust so that we're aligned within the uStreamer.

Sure. v0.1.1 just released after I updated the version. I believe we'll have v0.1.2 soon :laughing:

Cargo.toml
```diff
@@ -48,7 +48,7 @@ rand = "0.8.5"
 tokio = { version = "1.35.1", default-features = false }
 tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
-up-rust = { git = "https://github.com/eclipse-uprotocol/up-rust", rev = "3a50104421a801d52e1d9c68979db54c013ce43d" }
+up-rust = "0.1.1"
```
PLeVasseur, line 51: :tada:

src/rpc.rs
```diff
@@ -12,80 +12,125 @@
...
-use up_rust::{RpcClient, RpcClientResult, UAttributesError, UMessage, UMessageError, UUri};
+use up_rust::{
+    communication::{CallOptions, RpcClient, ServiceInvocationError, UPayload},
+    LocalUriProvider, UAttributes, UCode, UMessageType, UPayloadFormat, UPriority, UStatus, UUri,
+};
 use zenoh::prelude::r#async::*;
 
+pub struct ZenohRpcClient {
+    transport: Arc<UPClientZenoh>,
+    uri_provider: Arc<dyn LocalUriProvider>,
```
PLeVasseur, line 25: This is interesting. I would have assumed (maybe incorrectly) that you'd have implemented `LocalUriProvider` on `UPClientZenoh`, given that holds the `authority_name`.

Am I missing something?

Tagging in @sophokles73 for his thoughts as I had planned to do so for `UPTransportVsomeip`.

src/lib.rs
```diff
@@ -96,14 +89,14 @@ impl UPClientZenoh {
...
-    /// let upclient = UPClientZenoh::new(zenoh_config::Config::default(), String::from("MyAuthName"))
+    /// let upclient = UPClientZenoh::new(zenoh_config::Config::default(), "MyAuthName")
     ///     .await
     ///     .unwrap();
     /// # }
     /// ```
     pub async fn new(
         config: zenoh_config::Config,
-        authority_name: String,
+        authority_name: impl Into<String>,
```
PLeVasseur, line 99: Did you run into a case where this was needed? Just my own curiosity :slightly_smiling_face:

src/rpc.rs
```diff
@@ -12,80 +12,125 @@
...
 use tracing::error;
-use up_rust::{RpcClient, RpcClientResult, UAttributesError, UMessage, UMessageError, UUri};
+use up_rust::{
+    communication::{CallOptions, RpcClient, ServiceInvocationError, UPayload},
+    LocalUriProvider, UAttributes, UCode, UMessageType, UPayloadFormat, UPriority, UStatus, UUri,
+};
 use zenoh::prelude::r#async::*;
 
+pub struct ZenohRpcClient {
+    transport: Arc<UPClientZenoh>,
```
PLeVasseur, line 24: Given that the `RpcClient` is implemented on a different struct `ZenohRpcClient` now, I think it's safe to rename `UPClientZenoh` to `UPTransportZenoh` since it truly only handles the uP-L1 uTransport APIs now.

src/utransport.rs
```diff
@@ -37,60 +37,28 @@ lazy_static! {
...
+fn invoke_block_callback(listener: &Arc<dyn UListener>, resp_msg: UMessage) {
+    match Handle::try_current() {
+        Ok(handle) => {
+            task::block_in_place(|| {
+                handle.block_on(listener.on_receive(resp_msg));
+            });
         }
-    }
+        Err(_) => {
+            TOKIO_RUNTIME
```
PLeVasseur, line 48: Following back up on this -- did you look into other solutions or is this the way to go about it when the code is running as a part of a different executor (async_std)?

src/rpc.rs
```diff
@@ -12,80 +12,125 @@
...
-        let source = *request.attributes.source.0.clone().ok_or_else(|| {
-            let msg = "attributes.source should not be empty".to_string();
-            error!("{msg}");
-            UMessageError::PayloadError(msg)
-        })?;
-        let zenoh_key = if let Some(sink) = request.attributes.sink.0.clone() {
-            self.to_zenoh_key_string(&source, Some(&sink))
-        } else {
-            self.to_zenoh_key_string(&source, None)
+impl RpcClient for ZenohRpcClient {
```
PLeVasseur, line 43: Would it suffice to use the `InMemoryRpcServer` from up-rust as our server? If not, why not?

Similarly, would the other default uP-L2 implementations from up-rust work? It's just the `RpcClient` that needs to be implemented differently for Zenoh?

src/rpc.rs
```diff
@@ -12,80 +12,125 @@
...
+    async fn invoke_method(
+        &self,
+        method: UUri,
+        call_options: CallOptions,
+        payload: Option<UPayload>,
+    ) -> Result<Option<UPayload>, ServiceInvocationError> {
+        // Get source UUri
+        let source_uri = self.uri_provider.get_source_uri();
+
+        let attributes = UAttributes {
```
PLeVasseur, line 53: Feels like this is ripe for a function in up-rust? Maybe on `UMessageBuilder`? WDYT?

PLeVasseur (commented):
Hey @evshary :wave: 

Left questions and comments to make sure we're aligned

src/rpc.rs
sophokles73: I had just recently removed the corresponding `UAttributes::request` function :laughing:

src/rpc.rs
```diff
@@ -12,80 +12,125 @@
...
+            type_: UMessageType::UMESSAGE_TYPE_REQUEST.into(),
+            id: call_options.message_id().into(),
```
sophokles73, line 55: this will only work if the client has actually provided a message ID. If not, then the transport is responsible for creating a UUID ...

src/rpc.rs
```diff
@@ -12,80 +12,125 @@
...
+            priority: call_options
+                .priority()
+                .unwrap_or(UPriority::UPRIORITY_UNSPECIFIED)
+                .into(),
+            source: Some(source_uri.clone()).into(),
+            sink: Some(method.clone()).into(),
+            ttl: Some(call_options.ttl()),
+            token: call_options.token(),
```
sophokles73, line 63: you need to put the payload format into the attributes as well ...

src/rpc.rs
PLeVasseur: Was there a reason? I must have missed this haha

src/rpc.rs
evshary: This is my original thoughts, too. However, in the [L2 API of up-rust](https://github.com/eclipse-uprotocol/up-rust/blob/946d2fa57e3f44df5c19aba994b911653cff6049/src/communication/in_memory_rpc_client.rs#L188), they take both `UPTransport` and `LocalUriProvider` as arguments, so I want to make them more consistent.
Do you think we can still put `LocalUriProvider` inside the `UPClientZenoh` and provide a different API here?

src/lib.rs
evshary: No, just found that I can make the code more general by accepting `impl Into<String>`. We can accept `String`, `&String`, `&str` here. Therefore, if you want have `UPClientZenoh::new()`, you don't always need `String::from("authority")`, but just only put `"authority"` here.

src/rpc.rs
evshary: Agree with you

src/utransport.rs
evshary: Not yet. It's a good topic worth more investigation, but I'll not put it in this PR.
In fact, I should keep the PR clean and do one thing at a time, but I've already put too many things in this PR now, which is not good practice :laughing:

src/rpc.rs
evshary: Yes, you're right. It suffices to use `InMemoryRpcServer`, or even the RpcServer in our example code.
In fact, `InMemoryRpcClient` still works in our case, but I just want to provide another efficient one for Zenoh users.

src/rpc.rs
evshary: Yes, I forgot that...

src/rpc.rs
evshary: Add it now.

src/rpc.rs
```diff
@@ -10,82 +10,131 @@
...
+            return Err(ServiceInvocationError::Internal(msg));
         };
 
         // Send the query
         let mut getbuilder = self
+            .transport
             .session
             .get(&zenoh_key)
-            .with_value(value)
+            .with_value(payload_data.as_ref())
```
sophokles73, line 95: Is there a penalty in invoking `with_value` with empty payload?

Cargo.toml
```diff
@@ -48,7 +48,7 @@ rand = "0.8.5"
...
 zenoh = { version = "0.11.0-rc.3", features = ["unstable"]}
```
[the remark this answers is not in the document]
evshary: Good catch, I forgot to update it!

src/rpc.rs
evshary: TBH, I don't know. But I agree that we can avoid invoking `with_value` while there is no payload.

src/rpc.rs
PLeVasseur: Not sure of the intent here. Pinging @sophokles73 to see if he can help clarify for us.

I thought we could just then pass in an Arc<> to `UPTransportZenoh` for both arguments.

WDYT?

src/rpc.rs
PLeVasseur: I'm not sure if you think this fits, then, but would it make sense to show that we are able to use the uP-L2 APIs from up-rust and feed in the UPTransportZenoh and everything works in the tests?

So for example, in addition to the tests you already have,

* UPTransportZenoh{configured as an RPC server} <-> ZenohRpcClient{holding UPTransportZenoh}
* UPTransportZenoh{configured as an RPC server} <-> UPTransportZenoh{configured as an RPC client}

You could add these:

* UPTransportZenoh{configured as an RPC server} <-> InMemoryRpcClient{holding UPTransportZenoh}
* InMemoryRpcServer{holding UPTransportZenoh} <-> UPTransportZenoh{configured as an RPC client}
* InMemoryRpcServer{holding UPTransportZenoh} <-> InMemoryRpcClient{holding UPTransportZenoh}
* InMemoryRpcServer{holding UPTransportZenoh} <-> ZenohRpcClient{holding UPTransportZenoh}

I guess this can work out to a lot of combinations in the test code. Perfect use case or test_case? Or bad idea altogether? :sweat_smile: 

Was just thinking that it'd be good to ensure that UPTransportZenoh fits these roles in the automated testing.

evshary: Hmm... I'm not sure whether these L2 APIs should be tested in the transport layer. In theory, we should only ensure the L1 API works here and don't need to care about how others use L1 API, which will be too many cases.
Furthermore, I'm wondering if it's worth to provide our own L2 RpcClient here. Perhaps users want more consistent API usage among different transport implementations.

src/rpc.rs
evshary: What I'm concerned about is how users think while using L2 API. They need to initialize `UPTransportZenoh` with `LocalUriProvider`, and then they should put these two into the L2 structure. I'm not quite sure whether it makes sense or not. Although users don't need to have `LocalUriProvider` in `ZenohRpcClient`, they still face the same issues while using other L2 APIs.

sophokles73: If `ZenohRpcClient` needs access to local resource URIs, then it will need to have access to a `LocalUriProvider`. If `ZenohTransport` implements both `UTransport` and `LocalUriProvider`, then you can pass in two Arcs referring to the same `ZenohTransport` instance. If not, then you need to provide Arcs to two different things. Having two separate traits simply allows for more flexibility ...

src/rpc.rs
sophokles73: FMPOV you only need to make sure that you implement the Transport Layer API correctly. In the end, that's the whole purpose of having a layered approach ...
You _may_ want to provide a more efficient RpcClient implementation if it is worth the effort. In that case you should also verify that you have implemented the RpcClient API correctly (as we do in up-rust).

src/rpc.rs
PLeVasseur: Do you think the additional flexibility is needed in this case @evshary? Or should it be baked into `UPTransportZenoh`? I lean toward the latter, but would like to hear your thoughts.

evshary: TBH, after @sophokles73's comments below, I feel like I want to take the L2 API away from up-transport-zenoh-rust. Then, it will not be an issue anymore.

The remaining question is, if we implement `LocalUriProvider` in `UPTransportZenoh`, users will use the L2 API like this way: `InMemoryRpcClient::new(Arc<UPTransportZenoh>, Arc<UPTransportZenoh>)`. @PLeVasseur How do you think about it?

src/rpc.rs
evshary: Then I might want to take the RpcClient away... It will keep the layer cleaner

src/rpc.rs
sophokles73: actually, I'd expect them to do
```rust
let transport = Arc::new(UPTransportZenoh::new());
let rpc_client = InMemoryRpcClient::new(transport.clone(), transport);
...
```
What is the alternative?

evshary: After a second thought, I think the better way will be `UPTransportZenoh::new(..., uri_provider: &dyn LocalUriProvider)`, and the `LocalUriProvider` can be reused in L2 API.

evshary: Oh, I didn't see @sophokles73's comments before refresh
I mean this way:
```rust
let uri_provider = LocalUriProvider::new();
let transport = Arc::new(UPTransportZenoh::new(&uri_provider));
let rpc_client = InMemoryRpcClient::new(transport, uri_provider);
...
```

sophokles73: what's wrong with that?

evshary: I don't prefer to have the duplicate argument in the function, but it's just my preference.

PLeVasseur: > TBH, after @sophokles73's comments below, I feel like I want to take the L2 API away from up-transport-zenoh-rust. Then, it will not be an issue anymore.

The bespoke implementation you made is more efficient, right? IMHO it'd be good to keep it so long as it works with the InMemoryRpcServer. :thinking: 

> The remaining question is, if we implement `LocalUriProvider` in `UPTransportZenoh`, users will use the L2 API like this way: `InMemoryRpcClient::new(Arc<UPTransportZenoh>, Arc<UPTransportZenoh>)`. @PLeVasseur How do you think about it?

I suppose this would be the 'default' way of going about things for most users, but they could impl LocalUriProvider if they wanted that flexibility.

sophokles73: > I don't prefer to have the duplicate argument in the function, but it's just my preference.

FMPOV there is no _duplicate argument_ since they have different type (and responsibility). The fact that you have a struct which implements both, is nothing more than an implementation detail, isn't it?

evshary: Ok, then majority rule :laughing:. I'll implement the trait in UPTransportZenoh.

evshary:
@sophokles73 @PLeVasseur I've implemented the `LocalUriProvider` for the `UPTransportZenoh`. Please let me know what you think about it.

src/lib.rs
```diff
@@ -75,79 +72,93 @@ pub struct UPClientZenoh {
...
-    /// Create `UPClientZenoh` by applying the Zenoh configuration, `UAuthority`.
+impl UPTransportZenoh {
+    /// Create `UPTransportZenoh` by applying the Zenoh configuration, local `UUri`.
     ///
     /// # Arguments
     ///
-    /// * `config` - Zenoh configuration. You can refer to [here](https://github.com/eclipse-zenoh/zenoh/blob/0.11.0-rc.3/DEFAULT_CONFIG.json5) for more configuration details.
-    /// * `authority_name` - The authority name. We need it to generate Zenoh key since authority might be omitted in `UUri`.
+    /// * `config` - Zenoh configuration. You can refer to [here](https://github.com/eclipse-zenoh/zenoh/blob/0.11.0/DEFAULT_CONFIG.json5) for more configuration details.
+    /// * `uri` - Local `UUri`. Note that the Authority of the `UUri` MUST be non-empty.
```
sophokles73, line 85: FMPOV it would add to transparency if you would document here that the resource ID must be 0 and then actually return an error if it isn't. Otherwise, a non-zero resource ID would be silently ignored and the user might wonder why ...

src/lib.rs
```diff
@@ -380,10 +368,10 @@ mod tests {
...
-    #[test_case("//192.168.1.100/10AB/3/0", Some("//*/FFFF/FF/FFFF"), Ok(MessageFlag::Request); "Broadcast Request Message")]
-    #[test_case("//192.168.1.101/20EF/4/B", Some("//*/FFFF/FF/FFFF"), Ok(MessageFlag::Response); "Broadcast Response Message")]
-    #[test_case("//192.168.1.100/10AB/3/80CD", Some("//*/FFFF/FF/FFFF"), Ok(MessageFlag::Notification); "Broadcast Notification Message")]
+    #[test_case("//*/FFFF/FF/FFFF", Some("//192.168.1.101/20EF/4/B"), Err(UStatus::fail_with_code(UCode::INTERNAL, "Wrong combination of source UUri and sink UUri")); "Impossible scenario 1")]
+    #[test_case("//192.168.1.100/10AB/3/0", Some("//*/FFFF/FF/FFFF"), Err(UStatus::fail_with_code(UCode::INTERNAL, "Wrong combination of source UUri and sink UUri")); "Impossible scenario 2")]
+    #[test_case("//192.168.1.101/20EF/4/B", Some("//*/FFFF/FF/FFFF"), Err(UStatus::fail_with_code(UCode::INTERNAL, "Wrong combination of source UUri and sink UUri")); "Impossible scenario 3")]
+    #[test_case("//192.168.1.100/10AB/3/80CD", Some("//*/FFFF/FF/FFFF"), Err(UStatus::fail_with_code(UCode::INTERNAL, "Wrong combination of source UUri and sink UUri")); "Impossible scenario 4")]
```
sophokles73, line 374: FMPOV comparing error messages makes your tests very volatile. The UCode will be more likely to remain stable than the actual message ...

evshary: Good idea. I wasn't aware that I could use UCode instead of UStatus. Indeed, the error message bothers me so much.

examples/rpc_client.rs
```diff
@@ -12,36 +12,70 @@
...
 
-    // invoke RPC method
-    println!("Send request from {src_uuri} to {sink_uuri}");
-    let result = rpc_client.invoke_method(sink_uuri, umsg).await;
-
-    // process the result
-    let payload = result.unwrap().payload.unwrap();
-    let value = payload.into_iter().map(|c| c as char).collect::<String>();
-    println!("Receive {value}");
+    while resp_listener.response.lock().unwrap().is_none() {
```
sophokles73, line 78: how about using [tokio::sync::Notify](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html) to notify the current thread about the message having been received?

src/rpc.rs
PLeVasseur: As I think both @sophokles73 and I were saying, if you find having a more efficient approach worthwhile, let's keep it.

But yes, that would imply needing to add tests as he wrote.

PLeVasseur (approved):
Overall only thing up in the air IMHO is whether to keep the RpcClient impl. Otherwise seems good.

examples/rpc_client.rs
evshary: Good to know this API.

evshary:
> Overall only thing up in the air IMHO is whether to keep the RpcClient impl. Otherwise seems good.

I'm still not sure whether keeping the `ZenohRpcClient` is a good idea or not, but I agree with you we can add some tests for it. I'll add some tests to ensure the `ZenohRpcClient` can communicate with `InMemoryRpcServer`. Let's put off the decision of whether to remove the API later.

sophokles73:
Or simply leave the RpcClient out for the time being. You can always add it later if you have given it more thought.

FMPOV it is more important to have a first version of the zenoh transport available (and on crates.io) that we can all start using ...

evshary:
OK, I have an initial test for `ZenohRpcClient` now. Ensure `ZenohRpcClient` can communicate with both L1 and L2 API.

sophokles73 (commented):
The only thing that a L1 implementation needs to make sure is that a message can be sent and delivered to a listener registered for a matching source UUri. That is seems to be verified by `tests/publish.rs`. The tests in `tests/rpc.rs` and `tests/ustreamer.rs` seem to be either redundant to the examples or the tests from up-rust. Maybe you want to consider omitting them and reducing the code base to be maintained ...

tests/l2_rpc.rs
```diff
@@ -0,0 +1,114 @@
...
+    // CY_TODO: Verify the what kind of resource_id we can listen to
+    rpc_server
+        .register_endpoint(
+            Some(&UUri::from_str("//*/FFFF/FF/0").unwrap()),
+            METHOD_RESOURCE_ID,
+            example_handler.clone(),
+        )
+        .await
+        .unwrap();
+    // Need some time for queryable to run
```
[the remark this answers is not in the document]
evshary: That means Zenoh queryable, which is used in RPC server.

evshary:
> The only thing that a L1 implementation needs to make sure is that a message can be sent and delivered to a listener registered for a matching source UUri. That is seems to be verified by `tests/publish.rs`. The tests in `tests/rpc.rs` and `tests/ustreamer.rs` seem to be either redundant to the examples or the tests from up-rust. Maybe you want to consider omitting them and reducing the code base to be maintained ...

For `tests/rpc.rs`, it mainly tests the RPC request and response message, which covers the different mechanisms in Zenoh for RPC. For `tests/ustreamer.rs`, it ensures that UPTransport can handle something uStreamer use case, which listens to different message types.

tests/l2_rpc.rs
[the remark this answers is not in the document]
sophokles73: My understanding is that UPZenohTransport declares a Zenoh Queryable for the endpoint registered by InMemoryRpcServer, right? And it might take a while to _announce_ the queryable across the Zenoh network? So to be sure that a request can be handled, the client needs to wait a little?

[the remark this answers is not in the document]
evshary: Yes, you're right.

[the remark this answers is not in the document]
sophokles73: Is there a way for the transport to know when the queryable has been announced on the network, e.g. by means of a callback or something like that? Waiting for an undefined amount of time does not look very deterministic to me ...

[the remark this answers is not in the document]
evshary: Hmm.. we don't have such a way to know this right now. But I think 1 second is pretty enough in this case. Let me loop @Mallets if you want to add any comments.

PR #58, opened 2024-07-26
Fix the Rust toolchain version at 1.74.1.

evshary:
Fix the nightly msrv [CI error](https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/actions/runs/10104128685/job/27942593912).
Also specify the Rust toolchain version to avoid the error in the future.


sophokles73 (approved):
LGTM

PR #65, opened 2024-08-01
Update the `Cargo.toml` and `Cargo.lock` for the release.

evshary:
* Sync the `Cargo.toml` with up-rust before release.
* As suggested in https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html, keep the changes `Cargo.lock` for debug in the future.


sophokles73 (approved):
LGTM

PR #71, opened 2024-08-07
Able to support passing config from arguments.

evshary:
Able to provide configuration in the arguments to increase the flexibility.

### DIFF
--- a/examples/common/mod.rs
+++ b/examples/common/mod.rs
@@ -10,20 +10,72 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
+use clap::Parser;
 use up_transport_zenoh::zenoh_config;
+
+#[derive(clap::ValueEnum, Clone, Copy, PartialEq, Eq, Hash, Debug)]
+pub enum WhatAmIType {
+    Peer,
+    Client,
+    Router,
+}
+
+#[derive(clap::Parser, Clone, PartialEq, Eq, Hash, Debug)]
+pub struct Args {
+    #[arg(short, long)]
+    /// A configuration file.
+    config: Option<String>,
+    #[arg(short, long)]
+    /// The Zenoh session mode [default: peer].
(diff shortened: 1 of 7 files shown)


examples/common/mod.rs
```diff
@@ -10,20 +10,72 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
+use clap::Parser;
 use up_transport_zenoh::zenoh_config;
 
+#[derive(clap::ValueEnum, Clone, Copy, PartialEq, Eq, Hash, Debug)]
+pub enum Wai {
```
PLeVasseur, line 17: `Wai` => `WhatAmI`, right? I can say for certain I would not have understand that if not for having worked with Zenoh a little. Maybe something more descriptive or leave a comment?

examples/common/mod.rs
```diff
@@ -10,20 +10,72 @@
...
+    /// Disable the multicast-based scouting mechanism.
+    no_multicast_scouting: bool,
+}
+
 #[allow(clippy::must_use_candidate, clippy::missing_panics_doc)]
 pub fn get_zenoh_config() -> zenoh_config::Config {
+    let args = Args::parse();
+
     // Load the config from file path
-    // Config Examples: https://github.com/eclipse-zenoh/zenoh/blob/0.10.1-rc/DEFAULT_CONFIG.json5
```
PLeVasseur, line 18: I like providing an example config file. Maybe we can add one to this repo instead of pointing to an external one?

PLeVasseur (approved):
Looks good. Consider adding an example config file.

examples/common/mod.rs
evshary: Agree with you. The reason I used `Wai` is to match the usage in Zenoh examples, but it's unclear as you said.

examples/common/mod.rs
evshary: Yes, I've added the config file.

PR #73, opened 2024-08-08
Update the CI workflow and add release workflow.

evshary:
Deal with https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/issues/66
"Stole" most of the flows from up-rust :laughing:


sophokles73 (approved):
LGTM

PR #74, opened 2024-08-08
Add footnote if the PC configuration prevents multicast connections

dmacattack:


evshary (approved):
I prefer a troubleshooting section instead of footnote, but maybe we can wait and create that until more tips for troubleshooting.

PR #76, opened 2024-08-12
Use the correct Rust version to fix MSRV issue.

evshary:
Now the CI kept failing. This is because the mismatch Rust tool version.


sophokles73 (approved):
LGTM

PR #80, opened 2024-09-30
Remove obsolete dependencies

sophokles73:
Removed crates from Cargo.toml which are not used anywhere in the code.
Also moved deps which are required for examples only to dev-dependencies block.


evshary (approved):
LGTM

Issue #79, opened 2024-08-16
Use pub/sub only of Zenoh (not queriables)


stevenhartley:
Per https://github.com/eclipse-uprotocol/up-spec/discussions/229, switch transport implementations of zenoh to use pub/sub only even for RPC and notification type of events. This is to drastically simplify the transport implementions

stevenhartley:
Related to the C++ issue https://github.com/eclipse-uprotocol/up-transport-zenoh-cpp/issues/109

evshary:
Done via https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/pull/83

PR #83, opened 2024-10-07
[#79] Do not use Zenoh Queryables for RPC

sophokles73:
Removed code implementing RPC based on Queryables. Adapted tests and
example code.


evshary:
Thanks for your update. I'll take a look later.
One thing I'm thinking is that should we keep the interoperability at the same release?
That is, do we need to wait for `up-transport-zenoh-python` update before merging this?

src/utransport.rs
```diff
@@ -63,158 +36,35 @@ fn spawn_nonblock_callback(listener: &Arc<dyn UListener>, listener_msg: UMessage
...
-        query
-            .reply(query.key_expr().clone(), payload)
             .attachment(attachment)
             .await
-            .map_err(|e| {
-                let msg = format!("Unable to reply with Zenoh: {e:?}");
-                error!("{msg}");
-                UStatus::fail_with_code(UCode::INTERNAL, msg)
-            })?;
+            .map_err(|_| UStatus::fail_with_code(UCode::INTERNAL, "Unable to send with Zenoh"))?;
```
PLeVasseur, line 62: I'd prefer to get the error message from Zenoh. I've been in circumstances where I needed on my fork for this repo to add _even more logging_ to understand what I goofed up.

PLeVasseur (approved):
LGTM!

I think @evshary's feedback is valuable before merging (if he has the time).

examples/l2_rpc_client.rs
```diff
@@ -30,7 +30,10 @@ async fn main() {
             .await
             .unwrap(),
     );
-    let rpc_client = Arc::new(ZenohRpcClient::new(zenoh_transport.clone()));
+    let rpc_client = InMemoryRpcClient::new(zenoh_transport.clone(), zenoh_transport.clone())
+        .await
+        .map(Arc::new)
```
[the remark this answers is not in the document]
sophokles73: Not sure if I understand what you intend to do. Can you propose a concrete code change?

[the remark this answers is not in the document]
PLeVasseur: Sure, here's what I mean in a [Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3c174ed0571038c4711aefc546659dc0).

Essentially by using .map() the .expect() gets a little "far away" from the cause of the error. It's a nit honestly, just something that stuck out.

If we had a series of .map() chained together, I can see the value of having an .expect() at the chain. Having .expect at the end also looks more aesthetically pleasing.

tests/l2_rpc.rs
evshary, line 1: Perhaps we don't need to verify L2 API in the tests if we decide to remove the L2 layer from up-transport-zenoh-rust.

tests/rpc.rs
```diff
@@ -144,12 +144,16 @@ async fn test_rpc_server_client(
         .register_listener(&src_filter, sink_filter.as_ref(), request_listener.clone())
         .await
         .unwrap();
-    // Need some time for queryable to run
+    // Need some time for listener to be registered and announced
     sleep(Duration::from_millis(1000)).await;
 
-    // Send Request with ZenohRpcClient (L2 API)
+    // Send Request with RpcClient (L2 API)
```
evshary, line 150: Same as above. We don't need to verify L2 API here.

examples/l2_rpc_client.rs
evshary, line 1: If there is no L2 API in up-transport, I guess we don't need to keep this here.

evshary (commented):
I only had a quick scan. Left some comments on whether we should keep the tests and examples for L2 API.
@PLeVasseur Big thanks for your review. It's always better to have more eyes on the changes.

examples/l2_rpc_client.rs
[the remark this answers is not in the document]
sophokles73: FMPOV the `expect` is _close enough_ to the cause of the error ...

tests/l2_rpc.rs
sophokles73: I thought about that as well. However, given that the L2 layer implementation in up-rust is only being tested with a mock transport, I still think that it is worthwhile to have some (rudimentary) tests verifying that the Zenoh transport works with the up-rust provided L2 API implementations. WDYT?

evshary: Then I'm fine with that. The remaining question is whether we should have L2 example inside up-transport. If yes, then I guess we should also have l2_rpc_server.rs, which is similar to the one in up-rust.

sophokles73: > The remaining question is whether we should have L2 example inside up-transport.

FMPOV it would be beneficial to have an example that illustrates how to create the UPZenohTransport but apart from that, the examples in up-rust should be sufficient, don't you think?

examples/l2_rpc_client.rs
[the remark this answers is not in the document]
PLeVasseur: Yeah, nbd from my point of view. I guess it just stuck out. Let's call it good. :v:

tests/l2_rpc.rs
evshary: I still prefer to make it consistent. If we have `l2_rpc_client.rs`, then it's clearer to also have `l2_rpc_server.rs`. Anyway, I don't think this is the blocking point. It can be added later.

evshary (approved):
LGTM. Let's merge it first and see what will happen.

Issue #78, opened 2024-08-16
Consider supporting associated `new` functions which take a `UUri` struct


PLeVasseur:
May just be me :sweat_smile: but I like using the `UUri::try_from_parts()` function from `up-rust` for creating `UUri`s.

To initialize a `UPTransportZenoh` I currently have to do that, then convert from the `UUri` back to a `String` or `&str`.

Can we consider supporting associated `new` functions which take a `UUri` struct to avoid that little dance?

evshary:
Fixed in https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/pull/84

PR #84, opened 2024-10-08
[#78] Support passing UUri into constructors

sophokles73:
The functions for creating a new UPTransportZenoh have been changed
to accept TryInto<UUri> which allows clients to pass in either
strings or UUri structs.


evshary (approved):
LGTM

PR #85, opened 2024-10-11
Pin to the older msrv version.

evshary:
Fix the CI failure
https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/actions/runs/11265771929/job/31328151458

If we use msrv 0.16.x, it needs at least Rust 1.78.0. I guess it will be good to support the minimum version we need.
Therefore, I decide to fix at 0.15.1.

### DIFF
--- a/.github/workflows/verify-msrv.yaml
+++ b/.github/workflows/verify-msrv.yaml
@@ -32,5 +32,5 @@ jobs:
         toolchain: ${{ env.RUST_TOOLCHAIN }}
     - name: check MSRV
       run: |
-        cargo install cargo-msrv --version 0.15.1
+        cargo install cargo-msrv
         cargo msrv verify
--- a/.github/workflows/verify-msrv.yaml
+++ b/.github/workflows/verify-msrv.yaml
@@ -32,5 +32,5 @@ jobs:
         toolchain: ${{ env.RUST_TOOLCHAIN }}
     - name: check MSRV
       run: |
-        cargo install cargo-msrv
-        cargo msrv --verify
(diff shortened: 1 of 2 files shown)


sophokles73:
IMHO the issue is that you are [explicitly setting the rust toolchain to 1.75](https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/actions/runs/11284682029/job/31386255175#step:3:91) using the `rust-toolchain.yaml` file.
FMPOV this is not really necessary. The msrv check in the nightly build is specifically designed to make sure that your code can actually be compiled using rust 1.75 ...
If you remove the `rust-toolchain.yaml` file, everthing should work as expected. Alternatively, you could explicitly switch to the `stable` toolchain before invoking `cargo install msrv-check` ...

evshary:
> IMHO the issue is that you are [explicitly setting the rust toolchain to 1.75](https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/actions/runs/11284682029/job/31386255175#step:3:91) using the `rust-toolchain.yaml` file. FMPOV this is not really necessary. The msrv check in the nightly build is specifically designed to make sure that your code can actually be compiled using rust 1.75 ... If you remove the `rust-toolchain.yaml` file, everthing should work as expected. Alternatively, you could explicitly switch to the `stable` toolchain before invoking `cargo install msrv-check` ...

I remember the reason I added `rust-toolchain.yaml` is because Zenoh couldn't compile with the latest compiler at that time. Since this is not the case, I agree it's fine to remove it now.

sophokles73 (approved):
LGTM

PR #87, opened 2024-10-21
Bump Zenoh version to the official 1.0.0

evshary:
Since Zenoh is ready to release 1.0, have a quick update.
However, an issue was found here. It still needs some investigation.
https://github.com/eclipse-zenoh/zenoh/issues/1552

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,11 +31,11 @@ pedantic = "deny"
 #nursery = "deny"
 
 [dependencies]
-anyhow = "1.0.90"
+anyhow = "1.0.75"
 async-trait = "0.1"
-bytes = "1.8.0"
+bytes = "1.6.1"
 lazy_static = "1.4.0"
-protobuf = { version = "3.7.1" }
+protobuf = { version = "3.3" }
 tokio = { version = "1.35.1", default-features = false }
 tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
@@ -44,7 +44,7 @@ zenoh = { version = "=1.0.0", features = ["unstable", "internal"] }
 
 [dev-dependencies]
 chrono = "0.4.31"
-clap = { version = "4.5.20", features = ["derive"] }
+clap = { version = "4.4.11", features = ["derive"] }
 test-case = { version = "3.3" }
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,8 +40,7 @@ tokio = { version = "1.35.1", default-features = false }
 tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
 up-rust = "0.2.0"
-#zenoh = { version = "=1.0.0-beta.3", features = ["unstable", "internal"] }
-zenoh = { git = "https://github.com/eclipse-zenoh/zenoh.git", branch = "main", features = ["unstable", "internal"] }
+zenoh = { version = "=1.0.0", features = ["unstable", "internal"] }
 
 [dev-dependencies]
 chrono = "0.4.31"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,8 @@ tokio = { version = "1.35.1", default-features = false }
 tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
 up-rust = "0.2.0"
-zenoh = { version = "=1.0.0-beta.3", features = ["unstable", "internal"] }
+#zenoh = { version = "=1.0.0-beta.3", features = ["unstable", "internal"] }
+zenoh = { git = "https://github.com/eclipse-zenoh/zenoh.git", branch = "main", features = ["unstable", "internal"] }
 
 [dev-dependencies]
 chrono = "0.4.31"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,11 +31,11 @@ pedantic = "deny"
 #nursery = "deny"
 
 [dependencies]
-anyhow = "1.0.75"
+anyhow = "1.0.90"
 async-trait = "0.1"
-bytes = "1.6.1"
+bytes = "1.8.0"
 lazy_static = "1.4.0"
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -212,40 +212,46 @@ impl UPTransportZenoh {
...
-        {
-            if version != UATTRIBUTE_VERSION {
-                let msg =
-                    format!("UAttributes version is {version} (should be {UATTRIBUTE_VERSION})");
-                error!("{msg}");
-                return Err(UStatus::fail_with_code(UCode::INVALID_ARGUMENT, msg).into());
-            }
-        } else {
-            let msg = "Unable to get the UAttributes version".to_string();
+        let mut attachment_iter = attachment.slices();
```
sophokles73, line 223: I had also started migrating to 1.0.0-rc.2. Here's what I came up with:
```rust
    fn attachment_to_uattributes(attachment: &ZBytes) -> anyhow::Result<UAttributes> {
        if attachment.len() < 2 {
            return Err(UStatus::fail_with_code(
                UCode::INVALID_ARGUMENT,
                "message has no/invalid attachment",
            )
            .into());
        }

        let attachment_bytes = attachment.to_bytes();
        let ver = attachment_bytes[0];
        if ver != UATTRIBUTE_VERSION {
            let msg = format!(
                "Expected UAttributes version {UATTRIBUTE_VERSION} but found version {ver}"
            );
            error!("{msg}");
            return Err(UStatus::fail_with_code(UCode::INVALID_ARGUMENT, msg).into());
        }
        // Get the attributes
        let uattributes = UAttributes::parse_from_bytes(&attachment_bytes[1..])?;
        Ok(uattributes)
    }
```

evshary: Yes, I believe this looks better. But in our original design, we want to use the iterator to get multiple values out orderly.
https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/zenoh.adoc#23-uattribute-mapping
I'm fine with using pure bytes instead of multiple values, which might be simpler for other languages.
If you also agree with that, then we might need to update the spec and other language usage to keep the interoperability.

sophokles73: My understanding of the ZBytes type is that the `to_bytes` function simply concatenates the slices of data into one continuous vector, regardless of where the slices had originally been allocated in memory. The outcome should therefore be exactly the same as before, shouldn't it?

evshary: @sophokles73 Oh, I missed that. Then I think it can work perfectly. Since you've already had another version, do you prefer to create another PR, or it's fine for me to update this PR?
BTW, Zenoh will be released officially today (If nothing goes wrong :laughing: )

sophokles73: be my guest 😃 Good luck with the release!

sophokles73: looks like the 1.0.0 release has been successful :-) Are you going to update this PR accordingly?

evshary: Yes, I will update the PR tomorrow to fit the changes!

sophokles73 (commented):
are these newer versions actually needed, i.e. doesn't it compile with older versions? If that is not the case, then I would suggest to keep the old versions in order to allow for better interoperability with other crates ...

evshary:
> are these newer versions actually needed, i.e. doesn't it compile with older versions? If that is not the case, then I would suggest to keep the old versions in order to allow for better interoperability with other crates ...

Make sense. I reverted back the changes, but do we have any policy of maintaining the version of crates? We might need to update the crates someday, so it would be better to have a reference for other libraries to refer to. For example, everyone should pin the dependencies version to what `up-rust` uses.

sophokles73:
> For example, everyone should pin the dependencies version to what `up-rust` uses.

IMHO there is no need to pin concrete versions. Cargo usually does a good job of finding versions of deps that all crates can work with. That job is easier to do, if we only define minimum required versions of deps in our crates. But maybe I am not getting your point?

Cargo.toml
```diff
@@ -40,7 +40,7 @@ tokio = { version = "1.35.1", default-features = false }
 tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
 up-rust = "0.2.0"
-zenoh = { version = "=1.0.0-beta.3", features = ["unstable", "internal"] }
+zenoh = { version = "=1.0.0", features = ["unstable", "internal"] }
```
sophokles73, line 43: why do you want to pin this to exactly `1.0.0`? Given that Zenoh now uses semantic versioning, it should also work with `1.0.14` or `1.7.5`, shouldn't it?

evshary: Yes, I think it's fine to remove it now. The `=` was added due to the unstable changes in alpha / beta before.

evshary:
> > For example, everyone should pin the dependencies version to what `up-rust` uses.
> 
> IMHO there is no need to pin concrete versions. Cargo usually does a good job of finding versions of deps that all crates can work with. That job is easier to do, if we only define minimum required versions of deps in our crates. But maybe I am not getting your point?

I'm not sure, but wouldn't it be a good practice to keep the deps up to date? I thought you want all the crates deps to be updated at the same time, but it seems like your idea is to have the minimum version we need and cargo will find a suitable version for us.

Cargo.toml
evshary: Updated

Cargo.toml
```diff
@@ -40,7 +40,7 @@ tokio = { version = "1.35.1", default-features = false }
 tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
 up-rust = "0.2.0"
-zenoh = { version = "=1.0.0-beta.3", features = ["unstable", "internal"] }
+zenoh = { version = "1.0.0", features = ["unstable", "internal"] }
```
PLeVasseur, line 43: :tada:

PLeVasseur (approved):
Congratulations on the v1.0.0! :tada: 

LGTM

PR #91, opened 2024-11-15
Add features to hide Zenoh unstable

evshary:
Deal with https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/issues/90


PLeVasseur (approved):
LGTM!

Can you cut a release so that I can try to integrate this into up-streamer-rust?

evshary:
> IMHO it would be good to also document the supported feature(s) in the doc comments of lib.rs so that it is included on docs.rs
@sophokles73 Good idea

> Can you cut a release so that I can try to integrate this into up-streamer-rust?
@PLeVasseur Sure thing

PR #93, opened 2024-11-29
Bump up-rust to 0.3.0

evshary:


evshary:
One thing I would like to check is up-rust 0.2.0 and 0.3.0 are incompatible, right?
Therefore, I should also bump up-transport-zenoh-rust from 0.3.1 to 0.4.0 in this case.

src/lib.rs
```diff
@@ -164,7 +164,7 @@ impl UPTransportZenoh {
             uri.authority_name.clone()
         };
         // ue_id
-        let ue_id = if uri.has_wildcard_entity_id() {
+        let ue_id = if uri.has_wildcard_entity_type() {
```
PLeVasseur, line 167: I think you'd want to check for both `uri.has_wildcard_entity_type()` and `uri.has_wildcard_entity_instance()` as this will be checking if both the entity type and instance are wildcarded before making the ue_id wildcarded.

PLeVasseur: I suppose this might mean the tests and examples will need to also get updated to comply with this.

PLeVasseur (changes requested):
Hey @evshary -- could you check the comments I left? Seems like a minor change.

sophokles73:
> One thing I would like to check is up-rust 0.2.0 and 0.3.0 are incompatible, right? Therefore, I should also bump up-transport-zenoh-rust from 0.3.1 to 0.4.0 in this case.

Yes, there have been (very few) breaking changes between up-spec `1.6.0-alpha.3` and `1.6.0-alpha.4`.

src/lib.rs
sophokles73: FMPOV the key needs to be `*` if the entity type is `0xFFFF` to make sure that we match messages from/to arbitrary services. In this case the instance ID is irrelevant.

However, if just the instance ID is `0xFFFF`, then we would still need to make sure to _only_ match messages to/from the specific entity type.

Example: For an entity type of `0x1234` and instance type `0xFFFF` (wildcard), the key's entity fragment should thus probably _not_ just be `*` (which would match all entity types), but something like `*1234`.

@evshary is this possible and would this also match `1234`, i.e. an entity type with the default (omitted) instance id (`0000`)?

PLeVasseur: Hmm, I see what you mean @sophokles73. What do you say @evshary?

evshary: Okay, I understand what you mean, @sophokles73. To achieve what you describe above, we need to use `$*` instead of `*` in Zenoh. I've added the changes.
@PLeVasseur I also added some tests. Please let me know whether this is what you expect.

src/lib.rs
```diff
@@ -28,6 +28,8 @@ pub use zenoh::config as zenoh_config;
 use zenoh::internal::runtime::Runtime as ZRuntime;
 use zenoh::{bytes::ZBytes, pubsub::Subscriber, qos::Priority, Session};
 
+const WILDCARD_ENTITY_TYPE: u32 = 0x0000_FFFF;
```
sophokles73, line 31: we should probably make these constants public in up-rust ...

sophokles73:
@PLeVasseur you also need to approve since you requested changes ...

src/lib.rs
evshary: I'm also thinking about this...
Instead of exposing `WILDCARD_ENTITY_TYPE`, perhaps we can have a function to get the entity type directly?

sophokles73: Can you create a corresponding issue in up-rust?

evshary: Sure

PLeVasseur (approved):
LGTM!

PR #94, opened 2024-11-29
Avoid using zenoh-unstable in the examples

evshary:
I just found `insert_json5` is a better way in examples code to avoid using unstable configuration API.


PLeVasseur (approved):
Cool, makes sense to me.

Issue #96, opened 2025-02-05
Update to up-rust 0.4


sophokles73:

evshary:
I suppose this can be closed after https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/pull/98 is merged?

sophokles73:
closed via #98

PR #98, opened 2025-02-05
Update to up-rust 0.4

sophokles73:
addresses #96

Also adopted latest changes to CI jobs from up-rust

### DIFF
--- a/.github/workflows/verify-latest-deps.yaml
+++ b/.github/workflows/verify-latest-deps.yaml
@@ -11,28 +11,29 @@
 #  SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************/
 
-name: Check MSRV
+# Checks if the crate can be built using the latest stable toolchain with the latest
+# compatible dependencies.
+# The job definition has been inspired by the example given in the Rust Cargo book.
+
+name: Check latest dependencies
 
 on:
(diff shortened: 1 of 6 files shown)


.github/workflows/verify-latest-deps.yaml
```diff
@@ -0,0 +1,39 @@
+# ********************************************************************************
+#  Copyright (c) 2024 Contributors to the Eclipse Foundation
```
evshary, line 2: 2025?

evshary (approved):
LGTM anyway

.github/workflows/verify-latest-deps.yaml
sophokles73: The copyright header always contains the year _of inception_, i.e. the year that the file has been created originally. All subsequent changes can be derived from the git log, according to the Eclipse Foundation ...

sophokles73:
I would suggest to also incorporate the updates to the UUri -> Zenoh key mapping before doing the next minor release. WDYT @evshary ?

evshary:
> I would suggest to also incorporate the updates to the UUri -> Zenoh key mapping before doing the next minor release. WDYT @evshary ?

Agree with that.

PR #102, opened 2025-02-07
Bump to 0.5.0

evshary:
Since we have a breaking change on https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/pull/101, it would be better to release a new minor version.


sophokles73 (approved):
LGTM

Issue #103, opened 2025-02-14
Add references to OFT specitems to code base


sophokles73:
The Zenoh transport spec [now contains OFT spec items](https://github.com/eclipse-uprotocol/up-spec/pull/260) which need to be covered by the code base.

sophokles73:
The [nightly build currently fails](https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/actions/runs/13371456573/job/37340671911) due to uncovered requirements from the Zenoh Transport spec. This seems to be mainly due to the fact that there are no corresponding unit tests in the code base and IMHO the code base would need some refactoring in order to allow implementing them ...

@evshary WDYT?

evshary:
> The [nightly build currently fails](https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/actions/runs/13371456573/job/37340671911) due to uncovered requirements from the Zenoh Transport spec. This seems to be mainly due to the fact that there are no corresponding unit tests in the code base and IMHO the code base would need some refactoring in order to allow implementing them ...
> 
> [@evshary](https://github.com/evshary) WDYT?

I always support refactoring if necessary. Since I know little about OFT, would you mind elaborating more on which parts are missing and why we need to refactor to implement them?

sophokles73:
The requirements are being traced based on so called _specification items_. Those are defined in the Zenoh spec file in up-spec, which is included in up-transport-zenoh-rust by means of a Git submodule. The spec items have a property called _oft-needs_ which defines, what needs to be provided to cover the requirements. In most cases, _oft-needs_ has the value `impl,utest`, which means that the requirement needs to be implemented and verified by means of a unit test.
The fact that a spec item is being addressed (implemented, unit tested) is indicated in the source code by means of specially formatted comments, e.g.

```rust
// [impl->dsn~up-transport-zenoh-key-expr~1]
```

or

```rust
// [utest->dsn~up-transport-zenoh-key-expr~1]
```

Not all of the requirements that are defined in the spec are currently covered, that is why the corresponding check in the nightly build fails.

In particular, there are no unit tests for most of the requirements around the mapping of UMessages to Zenoh resources yet. We only have tests for the creation of Zenoh key expressions ...

Maybe I am mistaken and it is just a matter of adding the missing unit tests, but my feeling is that some of the code may need to be restructured to have appropriate API to build the tests on.

WDYT?

evshary:
[...]

PR #104, opened 2025-02-14
Add references to OFT spec items

sophokles73:
Also replaced local copies of GitHub Action artifacts with references to
reusable assets from eclipse-uprotocol/ci-cd.
Also included tracing report results in release artifacts.

This addresses #103


src/lib.rs
```diff
@@ -28,7 +28,7 @@ pub use zenoh::config as zenoh_config;
 use zenoh::internal::runtime::Runtime as ZRuntime;
 use zenoh::{bytes::ZBytes, pubsub::Subscriber, qos::Priority, Session};
 
-const UATTRIBUTE_VERSION: u8 = 1;
+const UPROTOCOL_MAJOR_VERSION: u8 = 1;
```
evshary, line 31: Why do we rename it to `UPROTOCOL_MAJOR_VERSION`?

src/lib.rs
```diff
@@ -296,6 +301,7 @@ mod tests {
     #[test_case("//*/FFFFFFFF/FF/FFFF", Some("//my-host2/CD/4/B"), "up/*/*/*/*/*/my-host2/CD/0/4/B"; "Receive all Requests")]
     #[test_case("//*/FFFFFFFF/FF/FFFF", Some("//[::1]/FFFFFFFF/FF/FFFF"), "up/*/*/*/*/*/[::1]/*/*/*/*"; "Receive all messages to a device")]
```
evshary, line 302: Do you want to rename UAuthority here to match the spec?

evshary (commented):
Left some comments.

sophokles73:
@evshary would you mind taking another look?

src/lib.rs
evshary: I'm not sure. Although it is unlikely to happen, do we need to bump the uProtocol version when we have a different way to transform UAttributes into Zenoh attachments?

sophokles73: IMHO we do, in order to keep things compatible. Currently, the Zenoh transport spec is part of the uProtocol spec. Thus, the expectation is that if a sender which uses the Zenoh transport that is based on uProtocol 1.x (or 1.6 for that matter), then a receiver using a Zenoh transport based on uProtocol 1.x can correctly receive UMessages.
If we change the Zenoh transport spec, e.g. changing the encoding of the UAttributes, in an incompatible way, then the UProtocol spec version needs to be bumped to a new major version, because that expectation would no longer be met ...

evshary (approved):
LGTM

PR #105, opened 2025-02-24
Run cargo deny as part of nightly build

sophokles73:
Also added corresponding configuration that allows the crate to pass
the checks.


evshary (approved):
LGTM

PR #106, opened 2025-03-07
Upgrade to ring 0.17.13

sophokles73:
This fixes https://rustsec.org/advisories/RUSTSEC-2025-0009


PLeVasseur (approved):
Good catch, thanks @sophokles73!

evshary (approved):
LGTM

PR #108, opened 2025-03-14
Bump to 0.6.0

evshary:
Upgrade `up-transport-zenoh-rust` to include ring fix and apply `up-rust` update


sophokles73 (approved):
LGTM
Maybe you can also update the `bytes` dependency to version `1.10.1` as used by `up-rust`

PR #110, opened 2025-03-17
Disable the OFT CI in the release flow.

evshary:
Deal with https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/issues/109


evshary:
@sophokles73 I only disable the one in release flow. Please let me know if you think the nightly should also be disabled.

PR #111, opened 2025-03-17
Remove the dependency of requirements-tracing.

evshary:
Forgot to remove the dependency in https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/pull/110

My bad. @sophokles73 Would you mind reviewing this again? Hopefully, I didn't miss anything this time...

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -59,20 +59,22 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # TODO: Disable the OFT CI for the time being.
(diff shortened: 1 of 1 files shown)


sophokles73 (changes requested):
lines 62-75 should also be commented out because they require the requirements check to have run ...

evshary:
Thank you for the reminder! I also disabled the one in `Collect quality artifacts`.
 [Here](https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/pull/111/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR121) is another thing I want to check. I think this is more related to up-spec, right?

sophokles73 (approved):

PR #112, opened 2025-03-18
Get the submodule (up-spec).

evshary:
Fix the failure that we can't get up-spec
https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/actions/runs/13914084354/job/38934123380
image


sophokles73 (approved):
LGTM

PR #115, opened 2025-04-01
Add workflow for checking latest spec compatibility

sophokles73:
Adopted the way that up-spec reads in the OpenFastTrace configuration
params from an env file in the local workspace. This makes it easier
to adapt the file patterns and tags according to the evolvement of
up-spec. The former way of using GitHub repo variables for that
purpose was rather static and did not allow to use different sets of
patterns on different branches in the repo.


evshary (approved):
LGTM

PR #119, opened 2025-06-25
Improve OFT spec item coverage

sophokles73:
Added missing references to the OFT spec items in the codebase.
Also added unit tests to ensure that the OFT spec items are covered.


evshary (approved):
LGTM. Sorry that I didn't have time to review OFT and address those issues...

Issue #120, opened 2025-06-26
Support limiting the number of listeners that can be registered


sophokles73:
The [Transport Layer specification](https://github.com/eclipse-uprotocol/up-spec/blob/v1.6.0-alpha.5/up-l1/README.adoc#33-registerlistener) requires implementors to provide means to configure an upper limit for the number of listeners that can be registered and to enforce that limit.

Currently, the Zenoh transport does not support this.

PR #121, opened 2025-06-27
Limit number of listeners that can be registered

sophokles73:
Also factored out the listener registry into its own module to
improve testability and maintainability of the code.

Addresses #120

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,6 +26,7 @@ use zenoh::internal::runtime::Runtime as ZRuntime;
 use zenoh::Session;
 
 const UPROTOCOL_MAJOR_VERSION: u8 = 1;
+const DEFAULT_MAX_LISTENERS: usize = 100;
 
 pub struct UPTransportZenoh {
     session: Arc<Session>,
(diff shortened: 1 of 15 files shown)


src/listener_registry.rs
```diff
@@ -0,0 +1,240 @@
...
+        {
+            Ok(subscriber) => {
+                // [impl->dsn~utransport-registerlistener-number-of-listeners~1]
+                // [impl->dsn~utransport-registerlistener-listener-reuse~1]
+                locked_subscribers.insert((zenoh_key, comparable_listener), subscriber);
+                Ok(())
+            }
+            Err(e) => {
+                let msg = "Failed to register listener";
+                info!("{msg}: {e}");
```
evshary, line 177: We can use `warn` or` error` here.

src/lib.rs
```diff
@@ -57,24 +40,33 @@ impl UPTransportZenoh {
...
+    ///
+    /// assert!(
+    ///     UPTransportZenoh::new(zenoh_config::Config::default(), "//MyAuthName/ABCD/1/0", 10)
     ///         .await
-    ///         .unwrap();
+    ///         .is_ok());
     /// # }
     /// ```
-    pub async fn new<U>(config: zenoh_config::Config, uri: U) -> Result<UPTransportZenoh, UStatus>
+    pub async fn new<U>(
```
evshary, line 65: I'm thinking that we can have another API (`pub async fn new_with_listeners_num`) and use a default value here to avoid the API changes.
In fact, a better way is to create a UPTransportZenohBuilder to accept the arguments. This will provide the flexibility to add arguments in the future.

sophokles73: I guess a builder would be nice, indeed. FMPOV we should not be too concerned about breaking the API at this early stage. I'll update the PR and add a builder ...

src/listener_registry.rs
sophokles73: I am always hesitant to use these high log levels because in most cases, the situation we have run into is actually totally expected and handled properly and we only want to _duly note_ (in the log) that this has actually happened.
In this case I wonder if not being able to declare the subscriber on the Zenoh session is actually an expected case or indicates a (fundamental) error with the Zenoh setup that we cannot recover from, i.e. which requires restarting the system or some other (manual?) resolving?

evshary: It's about how we define the log level. IMHO, when users call the function, they expect it to work. The warning messages (which have a different color) can easily help users identify why it fails.

sophokles73:
@evshary I have pushed some changes

evshary (approved):

Issue #122, opened 2025-07-01
Do not implement LocalUriProvider trait


sophokles73:
The Zenoh transport currently also implements the `up_rust::LocalUriProvider` trait and requires client code to pass a full UUri into the builder when creating a transport.
However, the transport itself only ever needs to know about the authority that it is running on (for cases where a UUri in a UMessage being sent does not contain an authority name). The other information is only needed to implement the RPC related functions of `LocalUriProvider`. Passing in this information seems redundant and also creates the impression that the transport instance would be specific to a particular uEntity using it (identified by the entity ID in the UUri), which it is not. There is no problem at all with sharing the same transport instance across multiple RpcServer and/or RpcClient instances.

With `up_rust::StaticUriProvider` we already have a generic implementation available, which can be used with the Communication Layer components like `up_rust::InMemoryRpcClient` and `up_rust::InMemoryRpcServer` where the entity ID, version and resource ID are relevant.

I therefore suggest to no longer let `UPTransportZenoh` implement `up_rust::LocalUriProvider` but instead simply require the local authority name to be passed into the builder.

evshary:
I don't have a strong opinion on this. If the upper layer provides a way to manage the Uri, then we don't need to do so in the transport layer.

In the original design, if we can determine the `key_expr` (including Uri information) beforehand, we can create a Zenoh publisher to avoid reusing `key_expr` every time, which can improve performance. Now that we generate the `key_expr` for every UMessage, it's okay to remove `LocalUriProvider` trait.

evshary:
Close via #123

PR #123, opened 2025-07-03
Do not implement LocalUriProvider

sophokles73:
UPZenohTransport has been changed to no longer implement the
LocalUriProvider trait. This seems reasonable to allow for the
same transport instance to be used by multiple service implementations
having different entity types.

Addresses #122


examples/l2_rpc_client.rs
```diff
@@ -33,47 +34,48 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // initiate logging
     UPTransportZenoh::try_init_log_from_env();
 
-    println!("uProtocol RPC client example");
-    let zenoh_transport = UPTransportZenoh::builder("//rpc_client/1/1/0")
-        .expect("invalid URI")
+    info!("uProtocol RPC client example");
```
evshary, line 37: I think it's fine to use `println!` in the example. We want to show the messages to users.

tests/ustreamer.rs
evshary, line 1: I guess you want to leave the responsibility of the test to uStreamer itself. I would like to know @PLeVasseur opinion on this.

tests/blocking_callback.rs
```diff
@@ -10,36 +10,32 @@
...
+        if let Some(mut pl) = msg.payload {
+            if !pl.is_empty() {
+                let delay_millis = pl.get_u32();
+                if delay_millis > 0 {
+                    sleep(Duration::from_millis(u64::from(delay_millis))).await;
+                }
+            }
         }
-        *self.recv_data.lock().unwrap() = value;
+        let _ = self.0.send(msg_id).await;
```
evshary, line 38: I guess it's fine to panic if unable to send the messages

evshary (commented):
Overall LGTM, but would like to know why you removed some tests. Do you think it's not `up-transport`'s responsibility to do those tests?

examples/l2_rpc_client.rs
sophokles73: Yeah, you're probably right. I'll change that back in the examples.

tests/ustreamer.rs
sophokles73: There is nothing specific to test for uStreamer, or am I mistaken? If the transport implements the L1 spec, then it will work in the streamer use case, right?

evshary: Well, this was added when @PLeVasseur was developing uStreamer. We can remove it if he doesn't mind.

evshary (approved):
I will merge it, but we can add the test back if necessary.

PR #125, opened 2025-07-08
Disable OFT CI in the release flow for the time being.

evshary:
We want to release v0.7.0 before the OFT update, so we need to bypass the check for the time being.
https://github.com/eclipse-uprotocol/up-transport-zenoh-rust/actions/runs/16136514409


evshary:
Close because it's replaced by #126

PR #126, opened 2025-07-08
Allow releasing with not all spec items covered

sophokles73:
The release workflow has been adapted to allow a release to be
created and published even if not all requirements are covered (yet),
according to OpenFastTrace.


sophokles73:
This is supposed to replace #125

evshary (approved):

PR #128, opened 2025-08-02
Bump the version of Zenoh, up-spec, and up-rust.

evshary:
While Zenoh, up-rust, and up-spec have an update, the PR will bump their version.


evshary:
@sophokles73 Reading the CI error log, should I do something like this?
https://github.com/eclipse-dash/dash-licenses?tab=readme-ov-file#automatic-ip-team-review-requests

Issue #127, opened 2025-08-01
Handling of incoming messages


sophokles73:
Currently, we dispatch the handling of each incoming message to a dedicated tokio runtime.
At first sight this seems to make sense. However, IMHO we should let users decide in the handler(s) that they register, if they actually want to spawn a task or handle the message on the current thread. Users need to run tokio anyway, so they shouldn't have any problem using `tokio::spawn` in their `UListener` implementation.

WDYT @evshary ?

sophokles73:
This would also provide for more consistent behavior with the [MQTT 5 transport](https://github.com/eclipse-uprotocol/up-transport-mqtt5-rust/issues/56)

sophokles73:
I do understand that. The question still is, if we want the user to decide how to deal with this (as we do in the MQTT 5 transport) or not. It seems to me that the developers of Zenoh had made the same decision as I want to propose we do: point out the potential for blocking the thread but leave it to the user how to deal with it :-)

evshary:
I slightly prefer to hide the complexity, but I'm also fine with making all the transport protocols follow a consistent pattern.

sophokles73:
I'll prepare a PR

PR #129, opened 2025-08-05
[#127] Process incoming messages on Zenoh's tokio runtime

sophokles73:
Changed the transport to process incoming messages on the tokio runtime
that Zenoh's callback is running on. This provides for full flexibility
regarding the configuration of the tokio runtime and allows for more
efficient use of resources in scenarios where the transport is used
on a constrained environment.

Doc comments and examples have been updated to reflect this change and
to provide guidance on how to implement listeners that process messages
without blocking the incoming message callback.


evshary (approved):
Thank you for the update and detailed documentation.

PR #131, opened 2025-08-06
Prepare v0.8.0 release

sophokles73:


README.md
```diff
@@ -52,8 +52,8 @@ The libraries need to be added to the `[dependencies]` section of the `Cargo.tom
 
 ```toml
 [dependencies]
-up-rust = { version = "0.6" }
-up-transport-zenoh = { version = "0.7" }
+up-rust = { version = "0.7" }
```
evshary, line 55: Perhaps we can use `cargo add` instead. Then we don't need to update the section every time.

README.md
```diff
@@ -1,6 +1,6 @@
 # Rust based Eclipse Zenoh&trade; Transport Library for Eclipse uProtocol&trade;
 
-This crate implements the Zenoh transport as specified in [uProtocol v1.6.0-alpha.6](https://github.com/eclipse-uprotocol/up-spec/tree/v1.6.0-alpha.6).
+This crate implements the Zenoh transport as specified in [uProtocol v1.6.0-alpha.6](https://github.com/eclipse-uprotocol/up-spec/blob/v1.6.0-alpha.6).
```
evshary, line 3: BTW, what is the difference between blob and tree? It seems that th e blob link will be redirected to tree every time.

sophokles73: Here's what CLaude Sonnet has to say about it:
> When linking to GitHub repositories, "blob" and "tree" path segments serve different purposes:
>
> **blob**: Used to link to a specific file and its contents
>
> Example: https://github.com/user/repo/blob/main/src/lib.rs
> Shows the file content with syntax highlighting
> Used when you want to reference a specific file
>
> **tree**: Used to link to a directory/folder structure
>
> Example: https://github.com/user/repo/tree/main/src
> Shows the directory listing and its contents
> Used when you want to reference a folder or browse multiple files
> Quick reference:
>
> Use `/blob/` when linking to files
> Use `/tree/` when linking to directories
> Both are followed by the branch/commit name (like main or master)

sophokles73: I will update the links accordingly ...

README.md
sophokles73: You mean replace it with something like
> In order to use the transport, run the following command in your workspace
> ```sh
> cargo add up-rust
> cargo add up-transport-zenoh
> ```

evshary: Yes

PR #132, opened 2025-08-07
Don't check the dependencies when a PR is merged into main.

evshary:
As discussed in https://github.com/eclipse-uprotocol/ci-cd/issues/22, we don't do another check when a PR is merged into main.
This can help avoid too many requests to the license verification server.


sophokles73 (approved):
LGTM

PR #138, opened 2025-11-11
Use enum to simplify the UPTransportZenohBuilder

evshary:
It will be easier to use an enum, and we don't need to declare the Zenoh default config explicitly.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -142,7 +142,7 @@ struct CommonProperties {
 
 pub enum BuilderState {
     Initial,
-    Config(zenoh_config::Config),
+    Config(Box<zenoh_config::Config>),
     ConfigPath(String),
     Session(Session),
 }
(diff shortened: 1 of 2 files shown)


sophokles73 (commented):
This basically removes the Type State pattern altogether, doesn't it? With your changes, you can now (again) set any properties on the builder :-(

evshary:
I guess I didn't get what you wanted to do before.
Do you mean, for example, when users use `with_config`, then they can't call `with_config_path` or `with_session` again?

sophokles73:
> I guess I didn't get what you wanted to do before. Do you mean, for example, when users use `with_config`, then they can't call `with_config_path` or `with_session` again?

exactly :-)
see https://cliffle.com/blog/rust-typestate/

evshary:
How about this? Keep using the Type State, but reduce the number of types by using enum. I think it should be simpler.

sophokles73 (commented):
Hmm, this seems to only make a tiny difference in code length but only as long as the ConfiguredBuilderState variants are basically immutable, i.e. there are no additional configuration properties supported on them. What I also liked about the previous approach was that we could implement the `build` function without `match` because we already now what we are dealing with ...

FMPOV this change does not really improve the code but only makes it less flexible and reduces encapsulation.

evshary:
I see. FMPOV, using match instead of implementing multiple traits, is more readable and easier to understand.
Although it might be less flexible, I don't foresee we will have too many changes here in the future.
Readability is what I would like to improve.

Anyway, I don't think this change really matters and needs a lot of discussion. I will close the PR.

